PR #8, opened 2017-03-03
The SendAsync method fails when Content-Length is > 2GB.

d4nsc4v4:
Even using the 'HEAD' verb, the HttpClient tries to allocate a buffer for the Content. Adding HttpCompletionOption.ResponseHeadersRead parameter tells the HttpClient not to do this.


Tyrrrz:
Thank you.
Does this also mean GetStreamAsync requires a fix too?

PR #10, opened 2017-03-06
Playlist support

dr-BEat:
Added a method that allows to extract all video ids in a youtube playlist.
The APIs used is the same that is also used by the mobile version of youtube at m.youtube.com .
The nice thing about it this that it works even for the special Watch Later playlist which is currently not supported in the official Youtube Data APIs.


Tyrrrz:
Thanks for contribution.
Is there a way to make this work without external dependencies? I'm trying to avoid adding dependencies unless it's absolutely required for some core function.

dr-BEat:
I managed to reimplement the parsing using just regular expressions.
It actually looks cleaner now. :)

YoutubeExplode/YoutubeClient.cs
```diff
@@ -186,6 +186,44 @@ public YoutubeClient()
...
+            while (nextUrl != null)
+            {
+                var responseContent = await _requestService.GetStringAsync(nextUrl).ConfigureAwait(false);
+                nextUrl = null;
+
+                var videoIdMatches = Regex.Matches(responseContent, @"""video_id"": ""([0-9a-zA-Z_\-]+)""");
+                var currentVideoIds = videoIdMatches.Cast<Match>()
+                                             .Where(m => m.Groups[1].Success)
+                                             .Select(m => m.Groups[1].Value)
+                                             .Distinct(); //VideoIds occur multiple times in the json data make sure we have not dupplicates
```
Tyrrrz: Wouldn't it make more sense to `.Distinct` the final `videoIds` list instead? It also covers edge cases when the same ids appear on different pages for whatever reason.

YoutubeExplodeTests/YoutubeClientPlaylistParsingTests.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.IO;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using YoutubeExplode.Services;
+
+namespace YoutubeExplode.Tests
+{
+    [TestClass]
+    public class YoutubeClientPlaylistParsingTests
```
Tyrrrz: Is there a good reason to use mocks here? The reason I opted for real-life integration tests was because youtube has a tendency to change stuff. Mocked unit tests basically only cover the parsing part, which is very basic to begin with. Integration tests, on the other hand, cover both the parsing and the integration part. They also are easier to maintain because you don't need to keep and update the fake response files.

YoutubeExplode/YoutubeClient.cs
```diff
@@ -186,6 +186,44 @@ public YoutubeClient()
...
+        /// </summary>
+        public async Task<List<string>> GetVideoIdsInPlaylistAsync(string playlistId)
+        {
+            var videoIds = new List<string>();
+
+            //We use a youtube api that is also used by the mobile version of the youbute page (m.youtube.com).
+            //For this to work we need to add &ajax=1&tsp=1&app=m to our playlist url
+            var baseUrl = $"https://m.youtube.com/playlist?list={playlistId}&ajax=1&tsp=1&app=m";
+            var nextUrl = baseUrl;
+            while (nextUrl != null)
```
Tyrrrz: What do you think of using recursion instead of a while loop? Is there a risk of stack overflow? I think it can make the code cleaner.

YoutubeExplode/YoutubeClient.cs
```diff
@@ -302,5 +340,56 @@ public static string ParseVideoId(string videoUrl)
...
+        /// <returns>Whether the execution was successful or not</returns>
+        public static bool TryParsePlaylistId(string playlistUrl, out string playlistId)
+        {
+            playlistId = default(string);
+
+            if (playlistUrl.IsBlank())
+                return false;
+
+            // https://www.youtube.com/playlist?list=PLOU2XLYxmsIJGErt5rrCqaSGTMyyqNt2H
+            var regularMatch = Regex.Match(playlistUrl, @"youtube\..+?/playlist\?.*?list=(.+?)(?:&|$)").Groups[1].Value;
```
Tyrrrz: This link appears to represent a play list too:

https://www.youtube.com/watch?v=b8m9zhNAgKs&list=PL9tY0BWXOZFuFEG_GtOBZ8-8wbkH-NVAr

YoutubeExplodeTests/YoutubeClientPlaylistParsingTests.cs
dr-BEat: I used a mock here because I wanted to avoid failing tests because the actual playlist was changed/deleted.
You are right about youtube changing things though.
I would leave this test but add an additional one to the intergration tests. I just have to find a playlist that does not change too often.
What do you think?

Tyrrrz: Yeah, I also try to use videos that are unlikely to change, like music videos.
Indeed, with unit tests you don't have to worry about that, but you do have to worry about updating the `PlaylistpageX.json` when youtube changes something. If you find a playlist that is very unlikely to change you nearly eliminate the first problem and don't produce the second.
I initially tried to mock stuff for VideoInfo as well, but quickly realized that there's no point in regards to a project like this one. Keeping both unit and integration tests doesn't offer much advantage over having just the integration tests, but takes a considerable amount of effort to maintain.

YoutubeExplode/YoutubeClient.cs
dr-BEat: I will move it down to the return.

YoutubeExplodeTests/YoutubeClientPlaylistParsingTests.cs
dr-BEat: I see your point. Ok, I will remove this and add a new test to the integration tests.

YoutubeExplode/YoutubeClient.cs
dr-BEat: A stack overflow is very unlikely. Since we do one call per 20 videos.
In my opinion the loop version is a bit cleaner, the recursive version would need 2 parameters for url + continuation token and maybe even an accumulator list + one public method as a nice facade.

Tyrrrz: Makes sense

YoutubeExplode/YoutubeClient.cs
```diff
@@ -186,6 +186,44 @@ public YoutubeClient()
...
+            //We use a youtube api that is also used by the mobile version of the youbute page (m.youtube.com).
+            //For this to work we need to add &ajax=1&tsp=1&app=m to our playlist url
+            var baseUrl = $"https://m.youtube.com/playlist?list={playlistId}&ajax=1&tsp=1&app=m";
+            var nextUrl = baseUrl;
+            while (nextUrl != null)
+            {
+                var responseContent = await _requestService.GetStringAsync(nextUrl).ConfigureAwait(false);
+                nextUrl = null;
+
+                var videoIdMatches = Regex.Matches(responseContent, @"""video_id"": ""([0-9a-zA-Z_\-]+)""");
```
dr-BEat: Done!

YoutubeExplode/YoutubeClient.cs
dr-BEat: Done!

YoutubeExplode/YoutubeClient.cs
```diff
@@ -186,6 +186,44 @@ public YoutubeClient()
...
 
         /// <summary>
+        /// Gets all videos in a given playlist.
+        /// This method will does one GET request per every 20 videos in the playlist.
+        /// 
+        /// This also works for special playlists like the Watch Later playlist at https://www.youtube.com/playlist?list=WL,
+        /// but the _requestServices has to mimic a logged in user.
+        /// Same for other non public playlists.
+        /// </summary>
+        public async Task<List<string>> GetVideoIdsInPlaylistAsync(string playlistId)
```
dr-BEat: Done!

Tyrrrz:
Looks good. I'll merge it tomorrow and refactor to bring it in line with the rest of the code.

dr-BEat:
Cool. Let me know if you have anymore comments or need help.

Tyrrrz:
I was also wondering what happens when you try to get info on non-existing (or private) playlist.

Tyrrrz:
Just refactored your last commit in

dr-BEat:
A private playlist will just return 0 videos.
Just looked at your refactorings, looks good.
But why did you reset the complete project history?

Tyrrrz:
Sorry about that, I fucked up in a git script and ended up accidentally duplicating every single commit, didn't notice until I pushed it. Decided it would be better to reset at that point. :\
I mentioned your contribution directly in a comment inside the GetPlaylistInfoAsync method.

dr-BEat:
Don't worry about my contribution, its just sad to see all old commits lost.

Tyrrrz:
Yeah, sometimes it's too easy with git to shoot yourself in the leg :(

dr-BEat:
Haha, that is true.

PR #35, opened 2017-07-01
CC Cultures

TheInsaneBro:
Since the supported cultures are also dependent of the current running operating system it's not that save to assume all possible cultures will be there. [Here](https://msdn.microsoft.com/en-us/library/cc233982.aspx) can be found a list of what and since when the different cultures are supported. I'm running on windows 7 still and thus I've got into troubles when one video got a CC with one of the following cultures:

- my
- ro-MD
- ku
- es-419
- eo
- ia
- mg
- om
- so
- fa-AF

But I think it's okay to leave it there, because they are supported by newer versions of windows.

But there are some unsolved problems. There are languages of cultures which have not been added so far. These include, but are not limited to:

- la (Latin, [Video](https://www.youtube.com/watch?v=tgnqyt53YbU))
- ie (Interlingue, [Video](https://www.youtube.com/watch?v=53BN9tTDoCw))
- und [(undetermined languages](https://en.wikipedia.org/wiki/ISO_639-3#Special_codes), [Video](https://www.youtube.com/watch?v=o0u4M6vppCI))

I also added a fix for Chinise `yi` → `ii`.

Then I added a `VideoRequiresPurchaseException` which is thrown when the video requires a purchase. If so, we have no access to most video details. [Example video](https://www.youtube.com/watch?v=p3dDcKOFXQg)

@Tyrrrz Can you please review the code and give a statement on how to proceed?


Tyrrrz:
Thank you for your investigation and the pull request!

I didn't realize that `CultureInfo` is actually dependent on executing OS, but that does make sense. However in worst case scenarios (when it can't find the correct culture) it creates an object like `Unknown Language (iso_code)`, right?

I can see the Latin working correctly though:

Interlingua is correctly recognized, but Interlingue isn't (what's the difference anyway?):

And unknown language seems to also work fine:


That test was ran on Windows 10.

Also what is the `yi` language? Can you give an example video?

YoutubeExplode/Exceptions/VideoRequiresPurchaseException.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace YoutubeExplode.Exceptions
+{
+    /// <summary>
+    /// Thrown when the video is a paid Youtube Red video since then we have no access to more video details
+    /// </summary>
+    public class VideoRequiresPurchaseException : Exception
+    {
+        /// <inheritdoc />
+        public VideoRequiresPurchaseException()
+            : base()
```
Tyrrrz: Let's add a message so the users can understand what went wrong:
`base("The video is a paid Youtube Red video and cannot be processed")`

YoutubeExplode/YoutubeClient.cs
```diff
@@ -206,6 +206,12 @@ public YoutubeClient()
                 throw new VideoNotAvailableException(errorCode, errorReason);
             }
 
+            // Check for paid content
+            if (videoInfoDic.ContainsKey("requires_purchase") && videoInfoDic.Get("requires_purchase") == "1")
```
Tyrrrz: You only need one check:
`if (videoInfoDic.GetOrDefault("requires_purchase") == "1")`

YoutubeExplode/YoutubeClient.cs
```diff
@@ -404,10 +410,17 @@ public YoutubeClient()
...
                     // Fix for Hebrew
                     if (lang == "iw") lang = "he";
-
-                    var culture = new CultureInfo(lang);
-                    var caption = new ClosedCaptionTrackInfo(url, culture, isAuto);
-                    captions.Add(caption);
+                    // Fix for Yi
+                    if (lang == "yi") lang = "ii";
+
+                    try {
```
Tyrrrz: Does it actually throw an exception? In my case it just returns an "Unknown Language (iso_code)" object.

Tyrrrz:
Also I think the Youtube red part of the pull request should be in a different branch, because it's not related to closed captions at all. Can you revert that part of the commits from this branch and make a pull request from a different branch?

YoutubeExplode/YoutubeClient.cs
TheInsaneBro: [This](https://social.msdn.microsoft.com/Forums/vstudio/en-US/70f01bf8-290b-43e3-a2ae-08870220c3c8/cultureinfogetcultureinfo-behavior-change-on-windows-10?forum=netfxbcl) states that it only will create such an `und`-object on windows 10 (or later, I guess) which I can confirm. It makes sense because the `und`-language isn't specified on earlier systems.

I first wanted to check if the current culture exists on the current system and then proceed depending on that, but `CultureInfo.GetCultures` is not in .Net Standard ([which also makes sense](https://github.com/dotnet/corefx/issues/1669)). That's why I've gone with try-catch.

TheInsaneBro:
Language yi [example Video](https://www.youtube.com/watch?v=zmPa--jmxrI) (Chinese (China))

Tyrrrz:
Does it throw an exception for you, when a language cannot be found for culture info?

TheInsaneBro:
Yes it does. The exception in the catch-block

Tyrrrz:
This is weird, mine doesn't. What are you testing with, one of the demos? Is it also OS specific, whether it throws an exception or not?

TheInsaneBro:
I commented your review of that code:

[This](https://social.msdn.microsoft.com/Forums/vstudio/en-US/70f01bf8-290b-43e3-a2ae-08870220c3c8/cultureinfogetcultureinfo-behavior-change-on-windows-10?forum=netfxbcl) states that it only will create such an `und`-object on windows 10 (or later, I guess) which I can confirm. It makes sense because the `und`-language isn't specified on earlier systems.

I first wanted to check if the current culture exists on the current system and then proceed depending on that, but `CultureInfo.GetCultures` is not in .Net Standard ([which also makes sense](https://github.com/dotnet/corefx/issues/1669)). That's why I've gone with try-catch.

Tyrrrz:
But then we have a problem where, depending on executing system, the caller will have different set of closed captions available, which doesn't really make sense. I think the best approach here is just to dump `CultureInfo` altogether. If I remember correctly, when querying the closed captions, it also has a property for full name (or english name) or something, we can just create our own `Language` object and store them there. What do you think?

TheInsaneBro:
This sounds like a reasonable solution. The full name is stored in the `n`-key of the caption dictionary.

Tyrrrz:
Let's do that then. Have the `Language` model consist of `string Code` and `string Name`.

TheInsaneBro:
So I guess we should also remove the fixes for `iw` and `yi` then?

Tyrrrz:
Yeah

TheInsaneBro:
I've just come over the fact that the language name provided by YouTube is in the language of the caller (in my case, all language names are in german). Do you think we should care about it? It doesn't really fit in the wpf demo, because everything else is english.

Tyrrrz:
What if we add `&gl=US` to the `get_video_info` request?

TheInsaneBro:
`&hl=en` works

Tyrrrz:
Perfect

TheInsaneBro:
I've implemented it. Now it works for me in all languages 👌

Tyrrrz:
Looks perfect. Ready for merge?

TheInsaneBro:
Yes it is

Tyrrrz:
Thanks a lot!

TheInsaneBro:
No problem. Thanks a lot for this awesome library 😉

PR #43, opened 2017-08-02
Moved Duration from VideoInfo to VideoInfoSnippet

SlowLogicBoy:
Since playlist also returns duration info moved it to Snippet.

* I wasn't able to run all test.

### DIFF
--- a/YoutubeExplode/Models/VideoInfo.cs
+++ b/YoutubeExplode/Models/VideoInfo.cs
@@ -67,7 +67,7 @@ namespace YoutubeExplode.Models
             long dislikeCount, bool isListed, bool isRatingAllowed, bool isMuted, bool isEmbeddingAllowed,
             IEnumerable<MixedStreamInfo> mixedStreams, IEnumerable<AudioStreamInfo> audioStreams,
             IEnumerable<VideoStreamInfo> videoStreams, IEnumerable<ClosedCaptionTrackInfo> closedCaptionTracks)
-            : base(id, title, description, keywords, viewCount, likeCount, dislikeCount, duration)
+            : base(id, title, duration, description, keywords, viewCount, likeCount, dislikeCount)
         {
             Author = author ?? throw new ArgumentNullException(nameof(author));
             Watermarks = watermarks?.ToArray() ?? throw new ArgumentNullException(nameof(watermarks));
--- a/YoutubeExplode/Models/VideoInfo.cs
+++ b/YoutubeExplode/Models/VideoInfo.cs
@@ -17,11 +17,6 @@ namespace YoutubeExplode.Models
         public ChannelInfo Author { get; }
 
         /// <summary>
-        /// Duration
-        /// </summary>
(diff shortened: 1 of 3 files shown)


SlowLogicBoy:
I also would like to expose UserId and AuthorName, but I would have to create another class I guess for playlist VideoInfo, or something like
Also maybe some people would like to get info this info too:
* Is Creative Commons Licensed (cc_license)
* Comment count (comments)
* Date Uploaded/Created (added or time_created)

Tyrrrz:
Which tests failed?

SlowLogicBoy:
Test that I was able to run passed, but Tests that require .csv datasource are crashing for me because I don't have Jet driver.

Tyrrrz:
Ah I see

Tyrrrz:
Thanks for your pull request!
I know playlist video info has access to more fields but it feels really confusing that `VideoInfoSnippet` essentially has more data than "full" `VideoInfo`.

TheInsaneBro:
Actually I'd like to see the Creative Commons Licensed property in there.

SlowLogicBoy:
@TheInsaneBro can you file an issue for this? so I or @Tyrrrz won't forget to implement this, unless you want to file PR yourself :)

Tyrrrz:
Yes please do. I also want to put those fields in models but I'm not sure what would be the best approach.
Maybe normal video info also has some kind of information about the license? I know youtube_dl extracts it somehow and i'm pretty sure it doesn't use playlist api.

SlowLogicBoy:
@Tyrrrz I only looked on what Playlist info returns, so I can't tell, but about your concern:

>I know playlist video info has access to more fields but it feels really confusing that VideoInfoSnippet essentially has more data than "full" VideoInfo.

I don't see the problem here since `VideoInfoSnippet` is a base model and it only makes `VideoInfo` easier to maintain :)

Tyrrrz:
> I don't see the problem here since VideoInfoSnippet is a base model and it only makes VideoInfo easier to maintain :)

Not sure I understood your point. Isn't that exactly why `VideoInfoSnippet` can't be a base model?

SlowLogicBoy:
I mean that VideoInfoSnippet should be treated as base model for info related to videos, and well VideoInfo just extends it. If we need to extend that data somehow differently we just create another model that extends VideoInfoSnippet.

Tyrrrz:
Ah I see what you mean now.
Maybe we can figure out good names for the models?

SlowLogicBoy:
VideoInfoSnippet -> VideoInfoModelBase; VideoInfoModel; VideoInfoBase
VideoInfo - I don't see problem with this name
PlaylistVideoInfo - Stuff that playlist returns on top of VideoInfoSnippet

With this we could do extension:
`PlaylistVideoInfo.ToVideoInfo()` (this would call `GetVideoInfoAsync()` to get full `VideoInfo`, if it would be possible since Models do not store `YoutubeClient`, maybe `new YoutubeClient()` isnt a bad thing)

PR #59, opened 2017-10-15
Views element is optional sometimes in playlist metadata

dr-BEat:
Some playlists like WatchLater (WL), maybe other private playlists as well, do not contain the views element in the metadata.
So I made it optional in GetPlaylistInfoAsync.


Tyrrrz:
Do it like this instead:

`(long?) playlistInfoXml.Element("views")?.Value ?? 0`

dr-BEat:
But Value is a string.
Do you mean:
`(long?)playlistInfoXml.Element("views") ?? 0`

Tyrrrz:
Yes, sorry. My bad.

dr-BEat:
Done! Thanks for the tip.

Tyrrrz:
Thanks!

PR #69, opened 2017-12-02
Added url properties to the video model

TheInsaneBro:
This can be useful in places where we e.g. get some videos over a playlist and want to use the video urls somewhere. I think the actual YouTube-library is the best place for such a conversion (from id to url).

Let me know what you think about it.


Tyrrrz:
Hi, thanks for PR. 

1. Can you move opening braces to new lines to be consistent with the rest of the code?
2. `Id` is never null by design
3. I think it would make more sense to move these to `Models.Extensions` class. I usually try to only keep integral properties inside model objects and move everything that can be derived to extensions.

Tyrrrz:
Thanks! Tell me when it's safe to merge.

TheInsaneBro:
It's ready :+1:

Tyrrrz:
Merged

PR #73, opened 2017-12-15
YouTube Live Streaming HLS Support

DominicMaas:
Expose the raw HLS playlist url (via `hlsvp`). This allows clients capable of playing .m3u8 files to play YouTube live streams.


Tyrrrz:
Once NuGet lists it, it should be available [here](https://www.nuget.org/packages/YoutubeExplode/4.1.0-alpha)

DominicMaas:
Thanks!

PR #84, opened 2018-01-13
added ability to download video in the highest available quality

TheBlubb14:
…and not capping at 720p

The new function "DownloadHighestQualityAsync" downloads the video stream with highest available quality and the best audio stream which is compatible with the video stream.
Then the two streams get muxed by FFmpeg.

You need FFmpeg.exe saved on your computer and provide the path to the method.
I wanted to provide the user with the video title and full download size, but I could not find a suitable way.

I added a commented out example in the demo console

### DIFF
--- a/DemoConsole/Program.cs
+++ b/DemoConsole/Program.cs
@@ -94,7 +94,7 @@ namespace DemoConsole
             id = NormalizeId(id);
 
             var progress = new Progress<double>(p => Console.Title = $"YoutubeExplode Demo [{p:P0}]");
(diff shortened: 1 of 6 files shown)


Tyrrrz:
Hi, thanks for the PR.

I've already had ideas about including FFMPEG in one form or another, but realized that it would be out of scope for this library and quite annoying to maintain. For example, [earlier I made a quick project](https://github.com/Tyrrrz/YoutubeDownloader) that demonstrates how to download streams to later multiplex them with FFMPEG.

I think it's best to create a separate nuget/library (e.g. `YoutubeExplode.FFMPEG` or something) that would do this, instead of putting it into core codebase.

Tyrrrz:
This is now done as part of https://github.com/Tyrrrz/YoutubeExplode.Converter

PR #96, opened 2018-02-13
Handle cases where caption duration is not specified

jerriep:
There are some instances where the duration ("d" attribute) is not specified in the captions. This PR adds a check for this condition and sets the `TimeSpan` to zero

Here are a couple of sample videos which causes this:
https://www.youtube.com/watch?v=5bm5tT0bN7c
https://www.youtube.com/watch?v=xc3Gl4rnWV4

The easiest way to test this is to open either of these URLs in the WPF sample app and then try and download the english caption.


Tyrrrz:
Nice find. Looks like captions that don't have duration also don't have text and as such don't bring any value. I say we just ignore such captions by validating that the text is not null or empty.

jerriep:
Sure, I'm heading off to bed but can update the check tomorrow

jerriep:
@Tyrrrz Hi Alexey, I changed as per your suggestion, and this also works on the sample videos linked to above.

Issue #100, opened 2018-03-28
Cannot write more bytes to the buffer than the configured maximum buffer size: 2147483647


Malediction:
Hello!

Since the change to just use HttpClient instead of IHttpService I'm getting errors when calling GetVideoMediaStreamInfosAsync() for this video:

https://www.youtube.com/watch?v=H1O_-JVbl_k

I think it's because you're not passing HttpCompletionOption.ResponseHeadersRead when calling SendAsync() on HttpClient (My own IHttpClient implementation did this), although I also have no idea why getting the media info for that video would want to allocate 2GB of buffer space.

I can send a PR if this seems like the logical way to fix this.

Tyrrrz:
Weird, the tests did pass. Please send PR if you can fix it.

PR #102, opened 2018-03-28
Set HttpCompletionOption.ResponseHeadersRead on all HTTP requests

Malediction:
Fixes https://github.com/Tyrrrz/YoutubeExplode/issues/100

Added unit test case for large video that fails when this header isn't specified.


Tyrrrz:
Beautiful. I wish all PRs were like this.

Issue #152, opened 2018-09-28
Playlist type that isn't supported (starts with OL) 


shaked6540:
https://www.youtube.com/playlist?list=OLAK5uy_lLeonUugocG5J0EUAEDmbskX4emejKwcM

I tried to parse the playlist id with `YoutubeClient.TryParsePlaylistId`, it returns false even though its a valid playlist (and public)
I've looked in the source code and it seems like you're checking for what the playlist ids start with, I guess there's a reason behind that but I'm wondering why this playlist type isn't supported, is the OL prefix something new that you didn't see yet, or is there a reason why its not supported?

Tyrrrz:
It's a new prefix. I don't understand what makes it different from others, looks like a regular playlist.

shaked6540:
Seems like this playlist type doesn't expose the author of the playlist and does not show the view count

PR #153, opened 2018-09-30
Added support for a new type of playlists

shaked6540:
(this is my first pull request please don't roast me)

This basically adds support for what I mentioned in #152 

Btw, this fails a unit test for a playlist with an ID which is supposed to be too long, but this new type of playlist has an id with the length of 41 characters and the test that suppose to fail has 35

### DIFF
--- a/YoutubeExplode.Tests/Data.cs
+++ b/YoutubeExplode.Tests/Data.cs
@@ -67,9 +67,9 @@ namespace YoutubeExplode.Tests
             yield return new TestCaseData("ULl6WWX-BgIiE"); // channel video mix
             yield return new TestCaseData("UUTMt7iMWa7jy0fNXIktwyLA"); // user uploads
             yield return new TestCaseData("PUTMt7iMWa7jy0fNXIktwyLA"); // popular user uploads
+            yield return new TestCaseData("OLAK5uy_lLeonUugocG5J0EUAEDmbskX4emejKwcM"); // music album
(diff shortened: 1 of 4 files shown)


Tyrrrz:
Hi and thanks!
Can you fix the tests so they are not failing?

Tyrrrz:
I still have doubts that the only defining feature of this playlist is that it doesn't have statistics. It looks like all videos inside of it are music videos but unfortunately I cannot watch them as they are unavailable for me.

Tyrrrz:
Do you know of any other examples that use this playlist type?

shaked6540:
No I haven't stumbled upon any of them, a user that uses my GUI wrapper for this library reported that it can't download this specific playlist. I'll try to get in touch with them

Tyrrrz:
Yeah this indeed looks like some autogenerated music playlist. Interesting. I do remember there was a PL type with a very long ID which was sort of like a mix playlist, based on an artist, while this one looks like a mix playlist based on an album.

Tyrrrz:
Also what is the project you're developing? I'm curious. 😁

Tyrrrz:
Okay, I've done some extra research and I'm tempted to rename `NormalWithoutStatistics` to `MusicAlbum`.

shaked6540:
The project I'm working on is https://github.com/shaked6540/YoutubePlaylistDownloader 
It's still not perfect but I'm trying to make it perfect slowly

And yeah I think `MusicAlbum` would fit more, I'll update it

Tyrrrz:
🎉

PR #154, opened 2018-10-03
Make the PLAYER_CONFIG parsing less brittle in GetVideoEmbedPageConfigAsync

mercenaryntx:
Hi,

I hope you don't mind, I extended the VideoUnavailableException with all the details collected about a video so they won't get lost when the GetVideoInfoAsync method throws an exception. So my app can still display basic info like title, thumbnail, etc. about a video, even when it cannot be processed.


Tyrrrz:
Hi,
Can you actually retrieve the video title if there's an error?

mercenaryntx:
> Hi,
> Can you actually retrieve the video title if there's an error?
> I know it might be tempting just to access video info but there's no guarantee it will always be there, or that any of the fields will be there in the future.

Sorry for the late reply, yes you can. 
Let me give you an example: https://www.youtube.com/watch?v=7MsP-87pACY

But it's true that there's no guarantee, that's why it should still throw an exception. 
But even if there was an error the retrieved data shouldn't be lost, so you can still work with what you have.

mercenaryntx:
Hi, sorry I wasn't aware that I cannot have multiple pull requests simultaneously, so my new commit automatically appeared under this pull request.

So, you must know already that something changed on the YouTube side, because now getting the player config throws an exception: 

Example: https://www.youtube.com/watch?v=sOexanZn-F4

Symptom: The JToken.Parse in GetVideoEmbedPageConfigAsync throws "Additional text encountered after finished reading JSON content: }. Path '', line 1, position 14985."

Reason: The embed code must have changed, there's no other parameter after the config JSON anymore. That .SubstringUntil(",'") part was not fail-safe enough.

So I introduced a .NET Regex using balancing groups to match JSONs. It's less efficient than having a recursive Regex, but unfortunately the built-in Regex class doesn't support recursion and the PCRE.NET Nuget package works with .NET Framework only. So that leaves us with the balancing group option only. I hope you don't mind.

Tyrrrz:
Can you please remove/revert "Extending the VideoUnavailableException with video details" from this branch? So that I can merge only the last commit.

Tyrrrz:
🎉

PR #162, opened 2018-10-10
Add seek support for streams

### LINKED ISSUE #161: Add Seek support for AsyncMultiStream
Edit:
[My current implementation that works](https://github.com/SlowLogicBoy/YoutubeExplode/blob/master/YoutubeExplode/Internal/YoutubeVideoStream.cs)
I also had to modify `MediaStream` so that all it's methods and properties pass to underlying stream.
How to call:
`return new MediaStream(info, new YoutubeVideoStream(_httpClient, info.Url, info.Size));`
[Instead all this](https://github.com/Tyrrrz/YoutubeExplode/blob/master/YoutubeExplode/YoutubeClient.MediaStream.cs#L20-L57)

Haven't tested it all, and I'm not sure if I used Count and position correctly.

With this I can seek video live in my asp.net core test application:
streamexample

Note:
* if you use `<video>` tag in html and try to seek video, it calls whole video url again (doesn't reuse Stream that you've already passed) and positions it with seek.
* Most of the time if you use `<video>` tag it request wider range of stream then `MaxSegmentSize` so there will not be many calls to youtube.

Downsides:
* ~~Possible more calls to youtube (HttpClient, and depends on buffer size, if it's > `MaxSegmentSize` it should be the same) therefore slower (might be a good idea to get max sized segment and use it until it's read fully)~~ Fixed in my PR

Upsides:
* Seekable stream (supports ranges in asp.net core)
* Downloads only what you need from youtube (no more no less).

SlowLogicBoy:
Fixes #161 
TODO:
- [x] Remove no longer used code.
- [x] Fix Failing tests
- [x] Cache Stream

~~Yay! I removed more then I added 😆.~~ 😿


YoutubeExplode/Internal/YoutubeVideoStream.cs
```diff
@@ -0,0 +1,91 @@
...
+                _currentStream.Dispose();
+                _currentStream = null;
+                bytesRead = await ReadAsync(buffer, offset, count, cancellationToken);
+            }
+            return bytesRead;
+        }
+        public override int Read(byte[] buffer, int offset, int count) =>
+            ReadAsync(buffer, offset, count).GetAwaiter().GetResult();
+
+        public override long Seek(long offset, SeekOrigin origin)
```
SlowLogicBoy: I'm not sure how Seeks work, I probably should throw instead of math'ing out what I can works with since user might get what's not expected

YoutubeExplode/Internal/YoutubeVideoStream.cs
```diff
@@ -0,0 +1,91 @@
...
+        public override void Flush() => throw new System.NotSupportedException();
+
+        public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            if (Position >= Length)
+                return 0;
+
+            count = Math.Min(MaxSegmentSize - 1, count);
+            if(_currentStream == null)
+                _currentStream = await _httpClient.GetStreamAsync(_url, Position, Position + MaxSegmentSize).ConfigureAwait(false);
```
SlowLogicBoy: I wonder could it be that `_url` might expire before everything is downloaded?

YoutubeExplode/Internal/YoutubeVideoStream.cs
SlowLogicBoy: [MemoryStream.Seek Implementation](https://referencesource.microsoft.com/#mscorlib/system/io/memorystream.cs,30f00da49c798fe1,references)

YoutubeExplode/Internal/YoutubeVideoStream.cs
SlowLogicBoy: I also noticed that before you would call `MaxSegmentSize - 1` I'm calling it here without `-1` and it seems to work ok, should I also add it or leave it as is right now?

Tyrrrz:
Oh I need a clear head to read through this. I'll try to get to this on the weekends. :)

YoutubeExplode/Internal/YoutubeVideoStream.cs
Tyrrrz: 1) No, leave it as is
2) 0 -> SegmentSize - 1; SegmentSize -> SegmentSize+SegmentSize - 1; etc..
This was done to split into even portions. In your case it doesn't matter since you're not splitting into anything beforehand. That said, technically your segments are 1 byte bigger than MaxSegmentSize so I would still add -1.

YoutubeExplode/Internal/YoutubeVideoStream.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            if (Position >= Length)
+                return 0;
+
+            if (_currentStream == null)
+                _currentStream = await _httpClient.GetStreamAsync(_url, Position, Position + MaxSegmentSize).ConfigureAwait(false);
+
+            var bytesRead = await _currentStream.ReadAsync(buffer, offset, count, cancellationToken).ConfigureAwait(false);
+            Position += bytesRead;
+            if (bytesRead == 0)
```
Tyrrrz: When is this condition met?

YoutubeExplode/Internal/YoutubeVideoStream.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        private const int MaxSegmentSize = 9_898_989;
+
+        private readonly string _url;
+
+        Stream _currentStream;
+
+        private long _position;
+
+        public YoutubeVideoStream(HttpClient httpClient, string url, long length)
```
Tyrrrz: It's fine to pass `MediaStreamInfo` instead of url/length

YoutubeExplode/YoutubeClient.MediaStream.cs
```diff
@@ -12,49 +12,10 @@ namespace YoutubeExplode
...
-                return new MediaStream(info, stream);
-            }
-            // If not rate-limited or not long enough - get the stream in one segment
-            else
-            {
-                // Get the whole stream as one segment
-                var stream = await _httpClient.GetStreamAsync(info.Url, 0, info.Size).ConfigureAwait(false);
-                return new MediaStream(info, stream);
-            }
+            return Task.FromResult(new MediaStream(info, new YoutubeVideoStream(_httpClient, info.Url, info.Size)));
```
Tyrrrz: I still think it would be nice if this was only used for rate-limited streams because otherwise it's a huge unnecessary overhead

Tyrrrz:
Added some comments

YoutubeExplode/Internal/YoutubeVideoStream.cs
```diff
@@ -0,0 +1,115 @@
+using System;
+using System.IO;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace YoutubeExplode.Internal
+{
+    internal class YoutubeVideoStream : Stream
```
Tyrrrz: Since it's not directly tied to Youtube in any way, can we call it `SegmentedHttpStream` or something?
It would also be nice if there was an extension method that would let me do `httpClient.GetSegmentedAsync(url)`.

YoutubeExplode/Internal/YoutubeVideoStream.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        Stream _currentStream;
+
+        private long _position;
+
+        public YoutubeVideoStream(HttpClient httpClient, string url, long length)
+        {
+            _url = url;
+            _httpClient = httpClient;
+            Length = length;
```
Tyrrrz: Is `Length` required? Can we avoid having to pass it so that it's more versatile?

YoutubeExplode/Internal/YoutubeVideoStream.cs
SlowLogicBoy: When current Stream from Youtube ends, just like with [AsyncMultiStream](https://github.com/Tyrrrz/YoutubeExplode/pull/162/files#diff-0d3b577d11ab6846312caf4545e1ebcbL41)

YoutubeExplode/YoutubeClient.MediaStream.cs
SlowLogicBoy: If I leave non rate-limited stream as it was, we won't be able to seek it, and I thought that additional overhead is not that big (additional class and couple if statements, which some of them I removed here), it will still call `_httpClient` once, as it did before (unless you seek it), the only difference (technically) will be that instead of calling `await _httpClient.GetStreamAsync(info.Url, 0, info.Size)` it will call `await _httpClient.GetStreamAsync(info.Url, 0, MaxSegmentSize)`. What do you think?

YoutubeExplode/Internal/YoutubeVideoStream.cs
SlowLogicBoy: Unfortunately we need that Length, otherwise I don't know how much youtube can give me, if I go over the boundaries `await _httpClient.GetStreamAsync(_url, Length, Length + MaxSegmentSize)` (basically when Position > Length) youtube returns 400 invalid range requested (unless you want me to catch that exception instead and say that stream is done).

I could try calculating if youtube doesn't return stream of size `MaxSegmentSize` as I request, I could say that this is the last stream they will give, but if MaxSegmentSize is exactly the size of a video, or if youtube decides to give me smaller stream then I requested anyway, it would just make it more brittle in my opinion.

YoutubeExplode/Internal/YoutubeVideoStream.cs
SlowLogicBoy: Renamed, and added an extensions.

YoutubeExplode/Internal/YoutubeVideoStream.cs
SlowLogicBoy: Added -1.

YoutubeExplode/Internal/YoutubeVideoStream.cs
SlowLogicBoy: And unfortunately response headers do not return overall size, only `Content-Length` which is for current request which in our case it's `MaxSegmentSize` most of the time for us.

YoutubeExplode/Internal/SegmentedHttpStream.cs
```diff
@@ -0,0 +1,115 @@
...
+        public override bool CanSeek => true;
+
+        public override bool CanWrite => false;
+
+        public override long Length { get; }
+
+        public override long Position
+        {
+            get => _position;
+            set
```
Tyrrrz: What will happen when a user sets `Position`? I'm guessing since the stream isn't re-queried, it's not going to work.

YoutubeExplode/Internal/YoutubeVideoStream.cs
Tyrrrz: Fair enough

YoutubeExplode/YoutubeClient.MediaStream.cs
Tyrrrz: I'm not sure I understand what do you mean by "it will still call `_httpClient` once". If the (non-throttled) stream has the length of 10 * MaxSegmentSize, won't that require 10 GET requests? I guess a simple solution would be to make segment size a constructor parameter and set it to `int.MaxValue` for non-throttled streams.

YoutubeExplode/Internal/SegmentedHttpStream.cs
SlowLogicBoy: You are completely right, my bad. Fixed.

YoutubeExplode/YoutubeClient.MediaStream.cs
SlowLogicBoy: Oh, I wasn't aware that non-throttled connection can have size bigger then MaxSegmentSize. I changed my code accordingly.

SlowLogicBoy:
> Hmm for some reason download time increased for
> `yield return new TestCaseData("H1O_-JVbl_k"); // very large video`
> in `Data.GetVideoIds` by significant amount after I added non rate limiting check.

My bad. Fixed.

YoutubeExplode/Internal/SegmentedHttpStream.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Threading.Tasks;
+
+namespace YoutubeExplode.Internal
+{
+    internal class SegmentedHttpStream : Stream
+    {
+        private readonly HttpClient _httpClient;
+
+        // Maximum segment size for rate limited streams
+        private const int MaxSegmentSize = 9_898_989; // this number was carefully devised through research
```
Tyrrrz: I would prefer if this was not a constant but rather passed as a value to constructor for throttled streams

YoutubeExplode/YoutubeClient.MediaStream.cs
```diff
@@ -12,49 +12,19 @@ namespace YoutubeExplode
...
+        public Task<MediaStream> GetMediaStreamAsync(MediaStreamInfo info)
         {
             info.GuardNotNull(nameof(info));
+            return Task.FromResult(new MediaStream(info, GetSegmentedStreamForMediaInfo(info)));
+        }
 
-            // Maximum segment stream size
-            const long segmentSize = 9_898_989; // this number was carefully devised through research
-
+        private Stream GetSegmentedStreamForMediaInfo(MediaStreamInfo info)
```
Tyrrrz: No need to make a separate method for this, just inlining is fine.
If you don't want to do `Task.FromResult` you can put `await Task.Yield().ConfigureAwait(false)` before it.

Tyrrrz:
Awesome! Final comments.

YoutubeExplode/Internal/SegmentedHttpStream.cs
SlowLogicBoy: 👍

YoutubeExplode/YoutubeClient.MediaStream.cs
SlowLogicBoy: I separated it, because I was doing:
```csharp
Stream stream;
if(isRateLimited)
    stream = ...;
else
    stream = ...;
```
and I don't like unassigned variables.
of course now I have code repetition which I also don't like :)

`await Task..` adds more overhead, I prefer doing `Task.FromResult()` where there is no async operations happening, instead of forcing `async` and `await`.

Tyrrrz:
Cool, tell me when this is ready for merge.

SlowLogicBoy:
If you looked at my latest `Code Cleanup` commit, then you can merge after appveyor completes it's build. ;)

PR #166, opened 2018-10-16
Fix finding player config in GetVideoEmbedPageAsync

### LINKED ISSUE #164: Newtonsoft.Json.JsonReaderException
Newtonsoft.Json.JsonReaderException: 'Additional text encountered after finished reading JSON content: }. Path '', line 1, position 14816.'

Video ID: wFQqbXk4W8Y and all other

It seems that youtube changed site content.

YoutubeClient.GetVideoEmbedPageConfigAsync cannot find expected string .SubstringUntil(",'");

Possible fix:

Instead

var part = raw.SubstringAfter("yt.setConfig({'PLAYER_CONFIG': ").SubstringUntil(",'");

use

var part = raw.SubstringAfter("yt.setConfig({'PLAYER_CONFIG': ").SubstringUntil("});writeEmbed()");

SlowLogicBoy:
Change "'," to "});" when taking embeded video config.

I decided to use `});` since it seems more appropriate, because it starts taking config from `({'PLAYER_CONFIG': ` so it's only natural to seek for closing brackets.

Fixes #164

### DIFF
--- a/YoutubeExplode/YoutubeClient.Video.cs
+++ b/YoutubeExplode/YoutubeClient.Video.cs
@@ -30,7 +30,7 @@ namespace YoutubeExplode
             // TODO: check if video is available
 
             var raw = await GetVideoEmbedPageRawAsync(videoId).ConfigureAwait(false);
-            var part = raw.SubstringAfter("yt.setConfig({'PLAYER_CONFIG': ").SubstringUntil(",'");
+            var part = raw.SubstringAfter("yt.setConfig({'PLAYER_CONFIG': ").SubstringUntil("});");
             return JToken.Parse(part);
         }
 


mercenaryntx:
That would be still too brittle. I already fixed this issue and ready to create a pull request with proper Regex. I just have to run a few personal errands first.

SlowLogicBoy:
Great! Can't wait to see your solution.

shaked6540:
I'm against regex, it's not necessary here unless you can make sure that it will always work and wouldn't slow it too much

And also as SlowLogicBoy said, it might be harder to understand and also maintain in case YouTube changes things again

mercenaryntx:
> I'm against regex, it's not necessary here unless you can make sure that it will always work and wouldn't slow it too much

It's always more reliable to match for exactly what you want than the rest of the string. 
The original solution was a bit negligent, even a JSON can contain a ",'" sequence. 
SlowLogicBoy's proposal is a bit better, however how can you be sure that this setConfig method is always going to be called with a one-property object? A few days ago it wasn't. Now the YouTube team made some noticeable changes.

But what do we know for certain? yt.setConfig is called with an object that has a PLAYER_CONFIG property which is also an object, a proper JSON. Matching a string with a pattern that describes these facts is hundred times more reliable than an obscure delimiter.

Speed is not an issue, really. Matching a single Regex won't make a difference, where are talking about insignificant microseconds here. I can make a performance test for you if you want, but honestly, establishing a network connection and getting the embed HTML has bigger performance overhead than this pretty simple Regex of mine.

> And also as SlowLogicBoy said, it might be harder to understand and also maintain in case YouTube changes things again

I highly doubt that YouTube is going to change the official standard of JSON anytime soon. :)

What I did here is simply the following: Tyrrrz's original string for matching the yt.setConfig call + a balancing group Regex to match the exactly same opening and closing curly braces. Meaning: while it's a valid JSON it's going to work.
Initially I wanted to put a PCRE there to define what a proper JSON is, but unfortunately .NET Regex doesn't support recursion and subroutines and the PCRE.NET library is not compatible with .NET Core/Standard.

It's true that the balancing group part is harder to understand, but believe me if it works now it won't need much of a maintenance later.

Tyrrrz:
I don't care about speed, regex performance is definitely the least of my worries right now.
I do agree that the original solution was brittle (hence why it broke), but I also liked that it was simple to understand. With regex there's a recurring problem that if someone else wrote a regex for something, it becomes hell to maintain it, unless you understand it perfectly.
However, I do like it if @mercenaryntx's regex is able to match the whole JSON block consistently. If it means I won't have to make updates as often when YouTube changes something, I'm willing to forego readability a bit in favor of regex approach.

Tyrrrz:
So what I'm going to do is merge this pull request to fix this ASAP and then look into @mercenaryntx's PR more thoroughly to avoid rushing things.

PR #170, opened 2018-10-17
 Added parsing username from user url

shaked6540:
This adds support for getting the username from a user URL.
`https://www.youtube.com/user/ProZD/` would return `ProZD`

### DIFF
--- a/YoutubeExplode/YoutubeClient.Channel.cs
+++ b/YoutubeExplode/YoutubeClient.Channel.cs
@@ -31,7 +31,7 @@ namespace YoutubeExplode
             username.GuardNotNull(nameof(username));
 
             if (!ValidateUsername(username))
-                throw new ParseException("Could not parse channel ID.");
+                throw new ArgumentException($"Invalid YouTube username [{username}].");
 
             // Get user page
             var userPage = await GetUserPageAsync(username).ConfigureAwait(false);
--- a/YoutubeExplode/YoutubeClient.Channel.cs
+++ b/YoutubeExplode/YoutubeClient.Channel.cs
@@ -30,6 +30,9 @@ namespace YoutubeExplode
         {
             username.GuardNotNull(nameof(username));
 
+            if (!ValidateUsername(username))
+                throw new ParseException("Could not parse channel ID.");
+
             // Get user page
             var userPage = await GetUserPageAsync(username).ConfigureAwait(false);
 
--- a/YoutubeExplode/YoutubeClient.Static.cs
+++ b/YoutubeExplode/YoutubeClient.Static.cs
@@ -216,5 +216,38 @@ namespace YoutubeExplode
                 ? result
                 : throw new FormatException($"Could not parse channel ID from given string [{channelUrl}].");
         }
+
+        /// <summary>
+        /// Tries to parses a username from a YouTube user URL
+        /// </summary>
+        public static bool TryParseUsername(string userUrl, out string username)
+        {
+            username = default(string);
+
+            if (userUrl.IsBlank())
+                return false;
+
+            var regularMatch = Regex.Match(userUrl, @"youtube\.com\/user\/([a-zA-Z0-9]{1,20})").Groups[1].Value;
+
+            if (regularMatch.IsNotBlank())
+            {
+                username = regularMatch;
+                return true;
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Parses username from a YouTube user URL.
+        /// </summary>
+        public static string ParseUsername(string userUrl)
+        {
+            userUrl.GuardNotNull(nameof(userUrl));
+
+            return TryParseUsername(userUrl, out string username)
+                ? username
+                : throw new FormatException($"Could not parse username from given string [{userUrl}]");
+        }
     }
 }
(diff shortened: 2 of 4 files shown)


shaked6540:
I messed up the regex a bit, I'll fix it

YoutubeExplode/YoutubeClient.Static.cs
```diff
@@ -216,5 +216,38 @@ public static string ParseChannelId(string channelUrl)
...
+        /// Tries to parses a username from a YouTube user URL
+        /// </summary>
+        public static bool TryParseUsername(string userUrl, out string username)
+        {
+            username = default(string);
+
+            if (userUrl.IsBlank())
+                return false;
+
+            var regularMatch = Regex.Match(userUrl, @"youtube\.com\/user\/([a-zA-Z0-9]{1,20})(?:\/|$)", RegexOptions.Multiline).Groups[1].Value;
```
Tyrrrz: 1) Why multiline?
2) I'd suggest to use `(.*?)(?:&|/|$)` instead of `([a-zA-Z0-9]{1,20})(?:\/|$)` just to make it simpler (there's no need to be very strict here), all validation logic should go to `ValidateUsername()` akin to `ValidateChannelId()`. Also remember to add `ValidateUsername()` to `GetChannelIdAsync()` method too later.

YoutubeExplode/YoutubeClient.Static.cs
```diff
@@ -216,5 +216,38 @@ public static string ParseChannelId(string channelUrl)
...
+        /// <summary>
+        /// Parses username from a YouTube user URL.
+        /// </summary>
+        public static string ParseUsername(string userUrl)
+        {
+            userUrl.GuardNotNull(nameof(userUrl));
+
+            return TryParseUsername(userUrl, out string username)
+                ? username
+                : throw new FormatException($"Could not parse username from given string [{userUrl}]");
```
Tyrrrz: Please put a period at the end of the exception message, for consistency.

YoutubeExplode/YoutubeClient.Static.cs
```diff
@@ -216,5 +216,38 @@ public static string ParseChannelId(string channelUrl)
...
+        public static bool TryParseUsername(string userUrl, out string username)
+        {
+            username = default(string);
+
+            if (userUrl.IsBlank())
+                return false;
+
+            var regularMatch = Regex.Match(userUrl, @"youtube\.com\/user\/([a-zA-Z0-9]{1,20})(?:\/|$)", RegexOptions.Multiline).Groups[1].Value;
+
+            if (regularMatch.IsNotBlank())
```
Tyrrrz: See above, add `&& ValidateUsername(regularMatch)`

YoutubeExplode/YoutubeClient.Static.cs
shaked6540: It needs to be multiline to work, otherwise, links without "/" at the end would not match

I'll add what you've asked for

YoutubeExplode/YoutubeClient.Channel.cs
```diff
@@ -30,6 +30,9 @@ public async Task<string> GetChannelIdAsync(string username)
         {
             username.GuardNotNull(nameof(username));
 
+            if (!ValidateUsername(username))
+                throw new ParseException("Could not parse channel ID.");
```
Tyrrrz: Please change the type to `ArgumentException` and format it as `Invalid YouTube username [{username}].` so it's consistent with other instances of this exception.

YoutubeExplode/YoutubeClient.Static.cs
shaked6540: Sorry, you're right. I checked with https://www.regextester.com/ and it showed no match unless I added the multiline flag, but I tried to execute it in c# and it seems to work even without it https://image.prntscr.com/image/W0BG11OiTEC0hb2l_fIGtQ.png

Tyrrrz:
And now also tests for `ValidateUsername()` 😆 

I'm sure you hate me by now

YoutubeExplode/YoutubeClient.Static.cs
Tyrrrz: I suggest using http://regexstorm.net/tester for testing .NET-flavored regex

shaked6540:
Don't worry, I need this library to work perfectly too so your comments are helpful

shaked6540:
Tests are a pain for me though, no matter how hard I try I can't run them in visual studio so they might fail. 
I'm always getting `System.InvalidOperationException: The following TestContainer was not found [path to a file that exists]`. no solution on google worked

Tyrrrz:
Hmm, that's weird. I'm able to run them locally on 2 different PCs.

Tyrrrz:
Right now they're failing because usernames `ABC`, etc, don't exist. Can you replace them with some random existing ones?

shaked6540:
Oh didn't see that comment, I just added a new test case data, do you want me to look for existing usernames to validate?

Tyrrrz:
Yeah it's better to have existing ones so it fits both the integration and unit tests and we don't have to make an extra method. 2-3 is enough.

Tyrrrz:
Thank you for your hard work :)

Issue #222, opened 2019-03-30
Poor performance in GetVideoAsync


SnGmng:
Since Yesterday my application was dead slow. I traced the issue down to the GetVideoAsync call. I updated YoutubeExplode yesterday, but i dont know if it is related to that update tho. I also benchmarked GetVideoAsync:
1139,06ms (average) per call!
You would have to wait 20s to load a 20 video playlist.
The GetPlaylistAsync takes reasonable 239,75ms (average) per call.

This is how i benchmarked:
`YoutubeClient client = new YoutubeClient();`
`Stopwatch stopwatch = new Stopwatch();`
`List<long> lstElapsed = new List<long>();`
`for (int i = 0; i < 100; i++)`
`{`
`    stopwatch.Reset();`
`    stopwatch.Start();`
`    var vid = await client.GetVideoAsync("-7fuHEEmEjs");`
`    stopwatch.Stop();`
`    var mil = stopwatch.ElapsedMilliseconds;`
`    lstElapsed.Add(mil);`
`    Debug.WriteLine("miliseconds elapsed: " + mil);`
`}`
`Debug.WriteLine("avg miliseconds elapsed: " + lstElapsed.Average());`

hope this gets fixed soon 😄

Tyrrrz:
Nothing changed in YoutubeExplode since yesterday, so if anything happened it was on YouTube's side. Also, `GetVideoAsync` is known to be pretty slow, hence why it was decoupled from the rest of the API in 4.0.

If you know how to improve the performance -- feel free to make a PR.

SnGmng:
Ive tested the previous versions and heres what i found:
4.6.7 averages 1139,06ms
4.6.6 averages 1009,46ms
4.6.5 averages 1046,5ms
4.6.4 averages 822,96ms
The time is the average of 30 GetVideoAsync calls.
Versions lower than 4.6.4 have about the same performance as 4.6.4.
In the last 3 versions there is something that slows down the calls by 300ms.
Allthough 300ms is a lot, it still doesnt explain why my application slowed down so much.
I will investigate that issue further.

Tyrrrz:
[...]

PR #223, opened 2019-03-30
Improve performance in description parser

SnGmng:
This increases the performance of description parsing by 100-200ms.
For more info see: https://github.com/Tyrrrz/YoutubeExplode/issues/222


Tyrrrz:
Do you know why is `.EndsWith` so much slower?

SnGmng:
I have no clue, i just tested it and it was faster 😆

SnGmng:
Im researching right now if a regex would be even faster

Tyrrrz:
I think you misprofiled, the longest executing part if `get_InnerText`
image

Tyrrrz:
This looks like the implementation in the version YoutubeExplode is using:
https://github.com/AngleSharp/AngleSharp/blob/f1c8bc6c6589b4b8a8cdc2c653fb65be8a977295/src/AngleSharp/Dom/Html/HtmlElement.cs#L517-561

Tyrrrz:
Switching from `.InnerText` to `.TextContent` halved the execution time to ~650ms with your benchmarks.

SnGmng:
Now ParseDescription is down to 18ms!
Allthough `.StartsWith` saved 200ms in my benchmarks, 
`TextContent` saved even more!

Here are my results for ParseDescription:
`.StartsWith` + `.InnerText`:  440ms
`.EndsWith` + `.InnerText`: 220ms
`.StartsWith` + `.TextContent`: 18ms
`.EndsWith` + `.TextContent`: 34ms

Tyrrrz:
Really curious why `.EndsWith` is so much slower, but I guess in the context of 18ms vs 34ms it doesn't matter very much. I blame `AngleSharp` for not providing clear distinction between `InnerText` and `TextContent`. Ideally I'd prefer a super simple HTML parser, but there are no alternatives.

Tyrrrz:
I'd still prefer `.EndsWith` though, because we're looking for truncated links and they end with triple dots

SnGmng:
Ok im fine with that. 10ms wont make a difference anyway

Issue #206, opened 2019-01-14
YoutubeExplode doesn't work with latest version of AngleSharp


darkdoctor:
Severity	Code	Description	Project	File	Line	Suppression State
Error		The "LinkAssemblies" task failed unexpectedly.
Mono.Linker.MarkException: Error processing method: 'System.DateTimeOffset YoutubeExplode.Internal.Parsers.VideoWatchPageParser::ParseUploadDate()' in assembly: 'YoutubeExplode.dll' ---> Mono.Cecil.ResolutionException: Failed to resolve AngleSharp.Dom.Html.IHtmlDocument
   at Mono.Linker.Steps.MarkStep.HandleUnresolvedType(TypeReference reference)
   at Mono.Linker.Steps.MarkStep.MarkType(TypeReference reference)
   at MonoDroid.Tuner.MonoDroidMarkStep.MarkType(TypeReference reference)
   at Mono.Linker.Steps.MarkStep.MarkField(FieldReference reference)
   at Mono.Linker.Steps.MarkStep.MarkInstruction(Instruction instruction)
   at Mono.Linker.Steps.MarkStep.MarkMethodBody(MethodBody body)
   at Mono.Linker.Steps.MarkStep.ProcessMethod(MethodDefinition method)
   at Mono.Linker.Steps.MarkStep.ProcessQueue()
   --- End of inner exception stack trace ---
   at Mono.Linker.Steps.MarkStep.ProcessQueue()
   at Mono.Linker.Steps.MarkStep.ProcessPrimaryQueue()
   at Mono.Linker.Steps.MarkStep.Process()
   at MonoDroid.Tuner.MonoDroidMarkStep.Process(LinkContext context)
   at Mono.Linker.Pipeline.Process(LinkContext context)
   at MonoDroid.Tuner.Linker.Process(LinkerOptions options, ILogger logger, LinkContext& context)
   at Xamarin.Android.Tasks.LinkAssemblies.Execute(DirectoryAssemblyResolver res)
   at Xamarin.Android.Tasks.LinkAssemblies.Execute()
   at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
   at Microsoft.Build.BackEnd.TaskBuilder.<ExecuteInstantiatedTask>d__26.MoveNext()	AnPlayer			

Hi, can you please let me know how I can fix this issue, this happened after i updated to the latest verison.

[...]

PR #244, opened 2019-05-25
Use latest AngleSharp for NetStandard1.3+

Tirraon:
See #206
I don't really like this solution with #if directive, but it seems impossible to do in other way.
And I don't know any other lightweight alternatives.

With this PR users on net45 and netstandard1.1 will reference latest possible for them version - 0.9.11.

Before this PR I couldn't use latest AngleSharp with latest YoutubeExplode.


Tyrrrz:
I'm seriously just considering writing my own lightweight read-only HTML parser to get rid of AngleSharp. Or maybe there's an alternative we can use instead?

Tyrrrz:
Clarification: I think your solution is really good considering the circumstences, I just hate that we have to do this :(

Tyrrrz:
Hi, just wanted to let you know I'm working on [my own parser](https://github.com/Tyrrrz/LtGt) to kill off dependency on AngleSharp. It needs a little effort and it should be ready soon.

Tirraon:
@Tyrrrz nice!
I will close this PR.

Tyrrrz:
Thanks for the PR though :)
Hopefully I will be able to add CSS selectors to LtGt soon and it should be enough.

PR #261, opened 2019-07-05
Decoupled URL object from youtubeclient

ivanprgomet-ip:


Tyrrrz:
Sorry for late reply. What benefit does this provide?

ivanprgomet-ip:
No problem man. Code testability/coverage is what I had in mind primarily. If it's not desired for the URL class i can remove my pr.

Tyrrrz:
I think it's fine because it's not public so we don't have to test it separately. I believe refactoring it into a separate class will decrease readability at this stage.

But thank you for the PR anyway!

PR #263, opened 2019-07-06
Replace keyword extraction regex

YoshiRulz:
No idea what the previous regex was supposed to do, `(?=\<q>...)` seems malformed. It was probably overkill anyway as you can't actually save the double quote character in a keyword (though you can enter it and it will remain until the page is reloaded). You might be able to replace the `IsNullOrWhiteSpace` check with `IsNullOrEmpty` as well, you can't enter just a space.

I checked this [online](https://regex101.com), not by running the test suite.

edit: also, this isn't a valid playlist ID ([try it](https://youtube.com/watch?v=x2ZRoWQ0grU&list=RDEMNJhLy4rECJ_fG8NL-joqsg), should be `RDx2ZRoWQ0grU`):
https://github.com/Tyrrrz/YoutubeExplode/blob/847bb23bccff609bdcd61ef9ff22d4cc2f70b3c2/YoutubeExplode.Tests/TestData.cs#L109-L110

### DIFF
--- a/YoutubeExplode/YoutubeClient.Playlist.cs
+++ b/YoutubeExplode/YoutubeClient.Playlist.cs
@@ -11,6 +11,9 @@ namespace YoutubeExplode
 {
     public partial class YoutubeClient
     {
+        /// <remarks>double quotes are silently stripped from keywords so it's safe to pair them</remarks>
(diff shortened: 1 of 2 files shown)


Tyrrrz:
IIRC the regex was matching paired quotes in order to extract keywords that contain spaces

PR #274, opened 2019-07-18
Use HtmlAgilityPack for net4.5 and netstandard2.0

SnGmng:
LtGt is a very new library and it is not as fast as libraries like HtmlAgilityPack or AngleSharp,
and because of this i think it would be better to use HtmlAgilityPack for .NET4.5 and .NET Standard 2.0 until LtGt is faster

### DIFF
--- a/YoutubeExplode/YoutubeClient.Channel.cs
+++ b/YoutubeExplode/YoutubeClient.Channel.cs
@@ -2,10 +2,16 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
-using LtGt;
-using LtGt.Models;
 using YoutubeExplode.Internal;
 using YoutubeExplode.Models;
+
+#if NETSTANDARD1_1
+using LtGt;
+using LtGt.Models;
+#else
+using HtmlAgilityPack;
+#endif
+
 
 namespace YoutubeExplode
 {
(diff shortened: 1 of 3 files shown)


Tyrrrz:
Do you have any examples where it performs slower? I benchmarked YoutubeExplode's methods with both AngleSharp and LtGt and didn't find any noticeable difference. HTTP calls take the most time so that parsing/lookup isn't really noticeable in comparison.

SnGmng:
I tested GetVideoAsync() since it is the slowest method:
My PR (HtmlAgilityPack) averages **762ms** (n=20)
Current release (LtGt) averages **1050ms** (n=20)

For me these 300ms is a noticeable difference, since my application calls GetVideoAsync() many times and this difference adds up, which makes my application noticeably slower.

Tyrrrz:
Hm, I'll take a look, I don't remember it being so slow.

SnGmng:
I just looked up, GetVideoAsync averages 822ms with AngleSharp (numbers from https://github.com/Tyrrrz/YoutubeExplode/issues/222)

Complete list:
- HtmlAgilityPack: **762ms**
- AngleSharp: **822ms**
- LtGt: **1050ms**

Tyrrrz:
I'll try to find out what's causing it. I'd rather make LtGt faster than use different libraries depending on target framework.

SnGmng:
My PR also replaces some linq with faster for-loops for example, [here](https://github.com/SnGmng/YoutubeExplode/blob/fa01fe98aaaae9e03066a8af57fbdd19566afb08/YoutubeExplode/YoutubeClient.Channel.cs#L91-L115)

Tyrrrz:
Just ran some benchmarks, you're right, it's on average 250ms slower on that request. I'll try to investigate.

Tyrrrz:
Are you blocked by this? Can you use your own fork in the meantime?

SnGmng:
Im not really blocked by this, it is just a small slowdown. And yes i can use my fork in the meantime.

SnGmng:
I am no longer maintaining this branch. I switched to my own library. If you want a faster youtube downloader, you can try it here: https://github.com/SnGmng/YoutubeDL-NET
But be aware that this is a early developmet version and still a bit unstable.

PR #290, opened 2019-08-19
Using decipherer functions instead of raw player source

omansak:


Tyrrrz:
Hi, thank you for the PR.
Can you please remove those few unrelated formatting changes in other methods?

Also, have you tested how often does this method work? Ultimately I'd prefer to replace the existing one rather than have 2 methods at the same time.

PR #291, opened 2019-08-20
Using decipherer functions instead of raw player source

omansak:


omansak:
Fixed formats

omansak:
What we want to do is very simple. Instead of regexing for the type of decrypt functions in the raw source, we just should to regex inside the function block. In this way, both in terms of improve regex performance and (as in the last update) we will avoid errors of similar names  in the future

Tyrrrz:
You don't have to create new PRs btw, you can just update your existing one. For future reference.

PR #313, opened 2019-10-20
Fix ValidatePlaylistId for My Mix

hig-dev:
"RDMM" is a special playlist id for the playlist "My Mix" from YouTube.


Tyrrrz:
Cool, thanks!

Issue #44, opened 2017-08-04
Get Channel Subscriptions and Playlists


SlowLogicBoy:
From channel id get it's subscriptions and playlists if publicly available.

`GetChannelSubscribtionsAsync(string channelId)`
returns `IEnumerable<string>` ChannelIds //at minimum maybe there would be more info available

`GetChannelPlaylistsAsync(string channelId)`
returns `IEnumerable<string>` PlaylistIds //at minimum maybe there would be more info available

Or maybe it is possible to get everything through `GetChannelInfoAsync(string channelId)`?

I haven't looked on how to do this. Worst case scenario would be to parse/scrape channel page.

https://www.youtube.com/channel/{channelId}/channels
https://www.youtube.com/channel/{channelId}/playlists

Tyrrrz:
Doesn't look like there's a way to get around parsing HTML :/

Tyrrrz:
I'm thinking maybe we should define a stricter scope of what YoutubeExplode should or shouldn't do. I'm not sure if I want it to turn into an alternative to Google API without the API key.

SlowLogicBoy:
Well, I can't really decide that, you do what you think is best, I could always write Extensions lib to do this parsing if I really need :)

Tyrrrz:
In that case, maybe you want to make it into a pull request? 😛

SlowLogicBoy:
We'll see, if I be in need of implementation for this, I will :)

Zyano:
I was wondering if this could be done without the API and it does appear to be possible and even relatively simple as well.

Youtube embeds a JSON object into the source of the playlist page which is then processed by polymer (the JS frontend) in order to show content.

Here is a gist from the information from the following page: https://www.youtube.com/user/LinusTechTips/playlists
https://gist.github.com/Zyano/857e5a4ad8df73e41b63aad6a71277b5 

The section in the gist is layered in the HTML like this.

` <script >
    window["ytInitialData"] = JSON SNIPPET/GIST GOES HERE
    window["ytInitialPlayerResponse"] = (null);
    if (window.ytcsi) {window.ytcsi.tick("pdr", null, '');}
  </script>`

It should be possible to parse this information to get the desired result.

Giggitybyte:
Any news or developments on this? This would be a neat thing to have.


As for your query a year ago @Tyrrrz, I might as well comment on that.
[...]

PR #332, opened 2020-01-08
WIP: Get playlists from channel/user page

Zyano:
This is still very much work in progress.
PR based on discussion from #44

A few things to note:
* This PR includes all the models used to parse the JSON from the channel/user page and the API used afterwards
* It attempts to cache the youutube api version used in the header to reduce the page load time after the first request by changing to the youtube ?pbj=1 and browse_ajax API
* It uses anglesharp and anglesharp.io which requires **.netstadnard2.0**
* A better public interfaces is still required


Tyrrrz:
198 files changed 😨

Zyano:
Yeah it's mostly the models required for parsing the entire youtube json model.

Ignore anything inside the folders Converter and YoutubeChannelPageModels.

Zyano:
I've gone a head and forced pushed the same changed but in a logic order now :)  This should make it easier to review and understand what has actually been done.

Tyrrrz:
I think it's way too much overhead for one feature, unfortunately.

Zyano:
Okay, in that case lets close the PR then

Tyrrrz:
Sorry about that. Thanks for the effort though.

Issue #259, opened 2019-06-28
GetChannelUploadsAsync Not Return Any Videos


hung1xpro:
Hello bro, nice to meet you but today i got new bug,
the GetChannelUploadsAsync not return any videos ...

Tyrrrz:
Doesn't work on all channels?

hung1xpro:
idk but i can't find channel can get video ...... all channel that i know still return 0

Tyrrrz:
Do you know if it stopped working today or was it after the last update? Can you check if it works on v4.7.3 instead of v4.7.4?

hung1xpro:
sorry cause i busy, i check ver 4.7.3 and still return 0 bro ... 
like that one: https://www.youtube.com/channel/UC0WP5P-ufpRfjbNrmOWwLBQ

Tyrrrz:
I see, seems YouTube broke something :(

hung1xpro:
so what  now bro ... we can't get that anymore ? :'(

Tyrrrz:
There might be a workaround, but I don't have time to look at it now :(

hung1xpro:
> There might be a workaround, but I don't have time to look at it now :(

oke bro, i can wait, just finish you work than check it late, btw thanks you for help :v

hung1xpro:
the thing is playlistId still work with list_ajax?..., then i think cause you code generate playlistid have some problem, that will help you if i right

Tyrrrz:
Looks like this was broken after #228 was fixed...

hung1xpro:
> Looks like this was broken after #228 was fixed...

Hmm now i can get all channel, but some don't like https://www.youtube.com/channel/UClyA28-01x4z60eWQ2kiNbA

I can get file json but u libary can't get, i think the code u use for get playlist have something wrong.

Tyrrrz:
It returns 80 videos for me, same as https://www.youtube.com/playlist?list=UUlyA28-01x4z60eWQ2kiNbA

hung1xpro:
But this 
` querry = YoutubeClient.ParseChannelId(querry);
                var channel = await _youtubeClient.GetChannelAsync(querry);
                var videos = await _youtubeClient.GetChannelUploadsAsync(querry);
                IReadOnlyList<YoutubeExplode.Models.Video> videoxx = videos;`
Still return 0 with videoxx.Count();

Tyrrrz:
Have you updated to latest version?

[...]

PR #338, opened 2020-02-08
Fix GetPlaylistAsync() only returning the first 200 videos for big playlists

### LINKED ISSUE #337: GetPlaylistAsync() only returning the first 200 videos for some big playlists
The method has a mechanism which advances the index, with the intention to requests all videos of a playlist. The problem is, by advancing the index only by 100 for each page, the returned videos for the first and second request will be the same, thus there are no new videos added and the method will return.

The first request (index=0) returns up to 200 videos (0-200), with index 100 returns the same 200 videos. A request with index 200 returns videos 101 to 300.

Solution: Instead of advancing the index by a fixed amount of 100, the countDelta variable can be used to increase the index. This results in all videos returned and we do not overshoot the playlist size (which would again result in only the first 200 videos in the request.)

polynoman:
…playlists.

Fixes #337

### DIFF
--- a/YoutubeExplode.Tests/TestData.cs
+++ b/YoutubeExplode.Tests/TestData.cs
@@ -121,6 +121,7 @@ namespace YoutubeExplode.Tests
             yield return new TestCaseData("PL601B2E69B03FAB9D"); // not sure
             yield return new TestCaseData("PLI5YfMzCfRtZ8eV576YoY3vIYrHjyVm_e"); // normal
             yield return new TestCaseData("PLWwAypAcFRgKFlxtLbn_u14zddtDJj3mk"); // large
+            yield return new TestCaseData("OLAK5uy_mtOdjCW76nDvf5yOzgcAVMYpJ5gcW5uKU"); // large alternative
             yield return new TestCaseData("RD1hu8-y6fKg0"); // video mix
             yield return new TestCaseData("RDMMU-ty-2B02VY"); // my mix
             yield return new TestCaseData("RDCLAK5uy_lf8okgl2ygD075nhnJVjlfhwp8NsUgEbs"); // music mix
--- a/YoutubeExplode/YoutubeClient.Playlist.cs
+++ b/YoutubeExplode/YoutubeClient.Playlist.cs
@@ -82,7 +82,7 @@ namespace YoutubeExplode
                     break;
 
                 // Advance index and page
-                index += 100;
+                index += countDelta;
                 page++;
             } while (page <= maxPages);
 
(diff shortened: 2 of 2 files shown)


Tyrrrz:
Hi, thanks for the PR.

I'm trying reaaaally hard to remember why I used 100 but I can't. I think it was because mix playlists worked differently or something? Not sure.

Could you check the playlists in test data (https://github.com/Tyrrrz/YoutubeExplode/blob/482c8707407ed4e8fb7d5fc7433539c2357653dd/YoutubeExplode.Tests/TestData.cs#L121-L131) and make sure that all of the videos are returned?

polynoman:
I just checked the playlists from your testcase and compared the results with the new version. 
The old version has no problem processing the "large" playlist of your examples, because for that playlists the index starts at 1. For the playlist "OLAK5uy_mtOdjCW76nDvf5yOzgcAVMYpJ5gcW5uKU" the index starts with zero, which will result in no new video in the second request thus, the stopping. You should add this playlist to the testcases.

For the "my mix" playlist, the return seems to be a bit random for both cases. Both will stop when there is no new video in the next request (countDelta = 0).

Tyrrrz:
Cool. So what I'm hearing is that the test playlists (except my mix) work correctly with your change? In that case, please add `OLAK5uy_mtOdjCW76nDvf5yOzgcAVMYpJ5gcW5uKU` to the test cases and I will merge it.

polynoman:
I just saw, that for the "large" the last few videos get lost. When the index starts with 1, adding `countDelta` will cause an overshoot in the end, resulting in some videos lost. By adding `countDelta -1` instead, we will not overshoot.

Is there a reason to start the index with 1 sometimes?

The index will be defined here:
`var index = playlistId.StartsWith("PL", StringComparison.OrdinalIgnoreCase) ? 101 : 0;`

If it is set to zero, adding just `countDelta` to the index will be fine.

Tyrrrz:
That changed seems to have been made to fix this issue: https://github.com/Tyrrrz/YoutubeExplode/issues/259

polynoman:
Ok, from what is see, the index is 101 when the playlist-code starts with PL, else it is 0.

For the channel case from #259 this computation will be 0 and it works fine.
If we use `countDelta` to advance the index, we can start with 0 for normal playlists as well, which then works fine for both the "large" playlist and my "large alternative"  `OLAK5uy_mtOdjCW76nDvf5yOzgcAVMYpJ5gcW5uKU` playlist.

So I recommend `var index = 0;` for all cases.

Tyrrrz:
Very well then :)
Please add your "large alternative" playlist to tests as well.

polynoman:
Done 👍

Tyrrrz:
Looks good.

I noticed that https://www.youtube.com/playlist?list=RDCLAK5uy_lf8okgl2ygD075nhnJVjlfhwp8NsUgEbs shows 91 videos but YoutubeExplode returns only the *first* 26 videos it seems. This happens on old version as well so it's a separate issue.

PR #381, opened 2020-05-21
Fix showing likes instead of dislikes

bcook254:
Noticed that `Engagement.ToString()` was showing the number of likes instead of dislikes.


Tyrrrz:
Thanks!

PR #387, opened 2020-05-30
Add maxPages parameter when searching for videos

PoLaKoSz:
# Description

This feature enable us to limit how many requests will be sent towards YouTube when searching for Videos.

The variable name is the same as in [v4.7.16](https://github.com/Tyrrrz/YoutubeExplode/blob/4.7.16/YoutubeExplode/YoutubeClient.Search.cs#L23) (`maxPages`) but -my opinion is that- `maxRequestCount` could be a better choice (i named it in the test to `maxSearchRequest`).

# Introduce breaking change?

- No

# How to use

For example if you want only the Video results from the first search
page, then set it to 1.
You don't have to specify it if you want to get as many Videos as
possible.

# Note about the test case

I think the test case is not fully accurate because i ran a few tests and the search endpoint sometimes returned 22, 40 and even 60 results per page, so i would like to ask (i never used library like this) what do You think about checking how many times the `YoutubeHttpClient.SendAsync()` method was hit (this is the last library method what it gets called before actually requesting from YouTube)?

### DIFF
--- a/YoutubeExplode.Tests/SearchSpecs.cs
+++ b/YoutubeExplode.Tests/SearchSpecs.cs
@@ -33,5 +33,20 @@ namespace YoutubeExplode.Tests
             videos.Should().NotBeEmpty();
             videos.Should().HaveCountLessOrEqualTo(maxVideoCount);
         }
+
+        [Fact]
+        public async Task I_can_search_for_YouTube_videos_and_limit_the_total_loaded_page()
+        {
+            // Arrange
+            const int maxSearchRequest = 1;
+            var youtube = new YoutubeClient();
+
+            // Act
+            var videos = await youtube.Search.GetVideosAsync("billie eilish", maxSearchRequest);
+
+            // Assert
+            videos.Should().NotBeEmpty();
+            videos.Should().HaveCountLessOrEqualTo(30);
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


maxkatz6:
@PoLaKoSz 
You can use Linq methods on IAsyncEnumerable with [System.Interactive.Async](https://www.nuget.org/packages/System.Interactive.Async/) package from https://github.com/dotnet/reactive.

Something like code below should help you:
```csharp
var firstPageItems = await youtube.Search.GetVideosAsync("billie eilish").Take(30).ToArrayAsync();
```

PoLaKoSz:
Thanks for the suggestion. There is already a [BufferAsync](https://github.com/Tyrrrz/YoutubeExplode/blob/5.0.5/YoutubeExplode/AccessibilityExtensions.cs) implementation which does similar what you wrote.

Lets assume this scenario:
- YouTube currently returns 20 results/page
- i hard code `.Take(20)`/`.BufferAsync(20)` to my code

Two weeks later:
- YouTube decide to return only 15 results/page
- i forgot to check this library so the library will do 2 requests which will return 30 results but i only be able to access the 20 results.
or
- YouTube decide to return 25 results/page
- i forgot to check this library so the library will return 25 results but i only be able to access the 20 results.

I assume that the YouTube search results are ordered by relevance. Probably i only need the first 4-5 video, but the library already sent one request why not save the remaining results too? (This is my opinion).
I came up with a better solution than this PR but i am gonna follow the proper etiquette and open an Issue first then i open a PR :smile:

Tyrrrz:
I'm wary of exposing implementation details such as pages/number of requests as semantics of  those may easily change as you pointed out.

PR #403, opened 2020-07-03
Added chapters to video results

btastic:
This PR will create support for getting chapters for a video.

Chapters, for those who don't know about them yet, are the small sections in the video's progress bar.

The implementation might not be the best, as I am relatively new to this library, but it works and it does not break any existing functionality.

Please let me know what you think, and how we can improve the code to fit this codebase better.

### DIFF
--- a/YoutubeExplode.Tests/ChapterSpecs.cs
+++ b/YoutubeExplode.Tests/ChapterSpecs.cs
@@ -1,0 +1,32 @@
+﻿using System.Linq;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Xunit;
+
+namespace YoutubeExplode.Tests
+{
+    public class ChapterSpecs
+    {
+        [Fact]
+        public async Task I_can_get_chapters_of_a_YouTube_video()
+        {
+            // Arrange
+            const string videoUrl = "https://www.youtube.com/watch?v=QBEItnHiqKM";
+            var youtube = new YoutubeClient();
+
+            // Act
+            var video = await youtube.Videos.GetAsync(videoUrl);
+
+            // Assert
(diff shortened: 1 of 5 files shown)


Tyrrrz:
Hi, thanks for the contrib.
I would generally recommend to create an issue first so that the feature can be discussed first.
I'm not sure I see value in adding chapters to the library. I find them very useful when watching the video itself, but not so much when downloading it. As far as I know, that's the primary use case for YoutubeExplode for most people.

btastic:
Hi, I was just doing a proof of concept for something I am personally interested in, that's why I tried and see if I could implement it.

Some longer music sets use this feature to display the current track playing. 

There is this almost ancient thing called ".cue" which is supported by some music players. When I download the video, convert it to mp3 I can generate a cue file that is used together with the mp3 file. I can now properly skip tracks with the ease of a click. Another plus is that I can see which track is currently playing.

Your comment sounds reasonable to have it not included in your library, however is there any way we could make this "InitialData" thing available whoever is interested in it?

Tyrrrz:
Thanks for clarification. I think it's better not to extend the scope of the library too much because that would make it more volatile to breakages.

Unfortunately exposing InitialData would mean taking dependency on that specific format, which is very likely to change in the future.

btastic:
Thats alright. I will do it differently. 

Just for the sake somebody stumbles on this: I will most likely access the `WatchPage` class via reflection to extract it from there.

olgrea:
Hello. I just wanted to provide a use-case for chapters when downloading a video. I'm using the library to download custom music mixes uploaded by users. Some uses chapters to timestamp sections of the mix where a song ends and the following one starts. Locally, I would use these timestamps along with FFmpeg to separate the mix into individual songs.

PR #417, opened 2020-07-16
Add search request limitation parameters

### LINKED ISSUE #389: Limit how many page should be loaded upon search
Lets assume this scenario:
- YouTube currently returns 20 results/page
- i hard code `.BufferAsync(20)` to my code

Two weeks later:
- YouTube decide to return only 15 results/page
- i forgot to check this library so the library will do 2 requests which will return 30 results but i only be able to access the 20 results.
or
- YouTube decide to return 25 results/page
- i forgot to check this library so the library will return 25 results but i only be able to access the 20 results.

I assume that the YouTube search results are ordered by relevance. Probably i only need the first 4-5 video, but the library already sent one request why not save the remaining results too? (This is my opinion).

What about this?
```c#
namespace YoutubeExplode.Search
{
    public class SearchClient
    {
        [Obsolete("Will be removed in the next major release (6.0)! Please use For() method instead.", false)]
        public async IAsyncEnumerable<Video> GetVideosAsync(string searchQuery) { ... }

        public SearchRequest For(string searchQuery) => new SearchRequest(_httpClient, searchQuery);
    }

    public class SearchRequest
    {
        private readonly YoutubeHttpClient _httpClient;
        private readonly string _searchQuery;

        public SearchResult(_httpClient, string searchQuery)

        public async IAsyncEnumerable<Video> GetVideosAsync() { ... }

        public async Task<IReadOnlyList<Video>> GetVideosAsyncAt(int pageNumber)
        {
            var response = await PlaylistResponse.GetSearchResultsAsync(_httpClient, searchQuery, page);

            ...
        }
    }
}
```
```c#
YoutubeClient youtube = new YoutubeClient();
SearchRequest searchRequest = youtube.Search.For("Tyrrrz");
IAsyncEnumerable<Video> allResults = searchRequest.GetVideosAsync();
IReadOnlyList<Video> only30Results = searchRequest.GetVideosAsync().BufferAsync(30);
IReadOnlyList<Video> resultsAtFirstPage = searchRequest.GetVideosAsyncAt(pageNumber: 1);
```

_**Code sample just a pseudocode**_

PoLaKoSz:
Set `firstPage` to 1 if you want to skip the first page results from the
search pages and get everything after it.
Set `takePage` to 2 if you want to get the search results only from two
pages.

Closes #389

### DIFF
--- a/YoutubeExplode.Tests/SearchSpecs.cs
+++ b/YoutubeExplode.Tests/SearchSpecs.cs
@@ -33,5 +33,19 @@ namespace YoutubeExplode.Tests
             videos.Should().NotBeEmpty();
             videos.Should().HaveCountLessOrEqualTo(maxVideoCount);
         }
+
+        [Fact]
+        public async Task I_can_search_for_YouTube_videos_and_get_only_the_third_page()
+        {
+            // Arrange
+            var youtube = new YoutubeClient();
+
+            // Act
+            var videos = await youtube.Search.GetVideosAsync("billie eilish", firstPage: 2, takePage: 1);
+
+            // Assert
+            videos.Should().NotBeEmpty();
+            videos.Should().HaveCountLessOrEqualTo(30);
+        }
     }
 }
--- a/YoutubeExplode/Search/SearchClient.cs
+++ b/YoutubeExplode/Search/SearchClient.cs
@@ -25,9 +25,18 @@ namespace YoutubeExplode.Search
         /// Enumerates videos returned by the specified search query.
         /// </summary>
         /// <param name="searchQuery">The term to look for.</param>
+        public IAsyncEnumerable<Video> GetVideosAsync(string searchQuery)
+        {
+            return GetVideosAsync(searchQuery, firstPage: 0, takePage: int.MaxValue);
+        }
+
+        /// <summary>
+        /// Enumerates videos returned by the specified search query.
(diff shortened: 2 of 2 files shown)


Tyrrrz:
Hi.
Can you make it as an overload instead of default parameters? Default parameters are, unfortunately, not backwards compatible.

PoLaKoSz:
What do You mean by "not backwards compatible"?
I didn't touched the test cases in the `SearchSpecs` file written before this PR and they are compiled successfully:
https://github.com/Tyrrrz/YoutubeExplode/blob/38e963cd18bc07f5615a8895bff81f16d1d38614/YoutubeExplode.Tests/SearchSpecs.cs#L16-L30

Tyrrrz:
They are source-compatible but not binary-compatible, so if another library was using YoutubeExplode and relied on that method with 1 parameter, it would suddenly no longer be there because there's only a method with 3 parameters. Default parameters are actually compiled client-side which makes them painful in the scenario where backwards compatibility is important.

PoLaKoSz:
Thanks You for the answer. I never thought about this so i leave here one of John Skeet's [article](https://codeblog.jonskeet.uk/2018/04/13/backward-compatibility-and-overloading/)

Tyrrrz:
Thanks!

PR #420, opened 2020-07-20
Fix UploadDate in PlaylistResponse

michalblaha:
Parsing from correct JSON property with fallback to previously used property

### DIFF
--- a/YoutubeExplode/ReverseEngineering/Responses/PlaylistResponse.cs
+++ b/YoutubeExplode/ReverseEngineering/Responses/PlaylistResponse.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Text.Json;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
+
 using YoutubeExplode.Exceptions;
 using YoutubeExplode.Internal;
 using YoutubeExplode.Internal.Extensions;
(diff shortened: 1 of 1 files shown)


Tyrrrz:
Is the old property not working on some playlists anymore?

michalblaha:
See attachment, for ex. last video in array. 
            **"added": "11/29/19",**
            "time_created": 1588575600,


And             **"time_created": 1588575600** means **GMT: Monday, 4 May 2020 07:00:00**




[yt.json.txt](https://github.com/Tyrrrz/YoutubeExplode/files/4948626/yt.json.txt)

Tyrrrz:
If I recall correctly, `added` is for when the video was added _to the playlist_, while `time_created` is when it was actually uploaded. It looks like `time_created` might be wrong due to some bug on their end, but `added` is wrong as well.

michalblaha:
From my point of view - date used in "added" is same as the date of the video on the YT website and is the same when video was really published.
 
Date in time_created is not visible anywhere else and IMHO this is the date when video was added to the playlist.

Tyrrrz:
Try to test it on this playlist: `PLI5YfMzCfRtZ8eV576YoY3vIYrHjyVm_e`.

Video:

image

Playlist Json:

image

The `added` date is September 9th, 2013, while `time_created` evaluates to `2013-09-30T18:10:23 UTC` which is closer to October 1st, especially if you consider timezones.

michalblaha:
In my case it's totally different. Most of videos in json has time_created same day, but they were uploaded many months earlier.

BTW, **if you ask for videoInfo for single video, date is OK**. The solution could be to remove these properties from PlaylistResponse…

Tyrrrz:
Which leads to my conclusion that both of these fields are actually wrong, at least sometimes.

These properties can't be removed because it shares the same model for `Video` as the other parts of the library. Ideally, we should change the way this data is scrapped so it uses more reliable sources instead of the iframe API, but until then it will have to stay like this.

michalblaha:
what about to use the older date from these two (until better solution or source of data)?

Tyrrrz:
In the example I mentioned, the older date (Sep 28) is actually less correct than the newer date (Sep 30).

PR #426, opened 2020-07-26
Improve performance in VideoClient.GetAsync

SnGmng:
Improves performance if you need both video metadata and the stream manifest.
Previously, the watch page would be fetched twice, now it is cached.
Without caching it would take 1400ms:
```csharp
var video = await youtube.Videos.GetAsync("B6N8-_rBTh8"); // 700ms
var manifest = await youtube.Videos.Streams.GetManifestAsync("B6N8-_rBTh8"); // 700ms
```
With caching it only takes 1000ms
```csharp
var video = await youtube.Videos.GetAsync("B6N8-_rBTh8"); // 700ms
var manifest = await youtube.Videos.Streams.GetManifestAsync(video); // 300ms
```

One more thing that is changed is that the user can now choose a "fast mode" (disabled by default):
```csharp
var video = await youtube.Videos.GetAsync("B6N8-_rBTh8", fastMode: true); // 150ms
```
The fast mode doent include like and dislike count, but is 4 times faster. 
Also caching is not possible when fast mode was enabled:
```csharp
var video = await youtube.Videos.GetAsync("B6N8-_rBTh8", fastMode: true); // 150ms
var manifest = await youtube.Videos.Streams.GetManifestAsync("B6N8-_rBTh8"); // 700ms
```

I really like your work, have a great day! 👍

### DIFF
--- a/YoutubeExplode.Tests/PlaylistSpecs.cs
+++ b/YoutubeExplode.Tests/PlaylistSpecs.cs
@@ -63,7 +63,7 @@ namespace YoutubeExplode.Tests
             var videos = await youtube.Playlists.GetVideosAsync(playlistUrl);
 
             // Assert
-            videos.Should().HaveCountGreaterOrEqualTo(20);
+            videos.Should().HaveCountGreaterOrEqualTo(19);
             videos.Select(v => v.Id.Value).Should().Contain(new[]
             {
                 "B6N8-_rBTh8",
@@ -72,7 +72,6 @@ namespace YoutubeExplode.Tests
                 "LsNPjFXIPT8",
                 "fXYPMPglYTs",
                 "AI7ULzgf8RU",
-                "VoGpvg3xXoE",
                 "Qzu-fTdjeFY"
             });
         }
--- a/YoutubeExplode.Tests/StreamsSpecs.cs
+++ b/YoutubeExplode.Tests/StreamsSpecs.cs
@@ -38,29 +38,6 @@ namespace YoutubeExplode.Tests
         }
 
         [Theory]
-        [InlineData("9bZkp7q19f0")] // very popular
-        [InlineData("SkRSXFQerZs")] // age restricted (embed allowed)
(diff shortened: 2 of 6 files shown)


SnGmng:
btw, the test failed because one of the videos in this playlist got deleted: https://www.youtube.com/playlist?list=PLr-IftNTIujSF-8tlGbZBQyGIT6TCF6Yd
and nice osu skills!

SnGmng:
fastmode has been removed and replaced with a better solution! It is now fast by default without drawbacks

Tyrrrz:
Hi, thanks for the PR.

The cache should be kept on the client object, rather than `Video` or `Playlist` because those are pure data models.  It can be a `Dictionary<VideoId, PlayerConfig>` or whatever.

You also need to consider expiration time because it shouldn't be cached forever. The data on the watch page may become out of date.

SnGmng:
I agree that this would be better, but I'm considering removing the caching again as it has become a rare side case in my last commit. 

In my latest commit i changed the GetVideo function to use a mixplaylist because it is faster.
https://github.com/Tyrrrz/YoutubeExplode/blob/88001564d5c3f90079a9af2aedf518dfaf2af265/YoutubeExplode/Videos/VideoClient.cs#L40-L96

The mixplaylist approach can even handle livestreams. Only unlisted videos still require the slower watchpage scraping approach. What are your thaughts about this?

Tyrrrz:
We can use the faster approach first, but separate the `try` and `catch` parts into different methods like I did with `GetManifestAsync`. As for caching, I would prefer to avoid it if possible as it opens a whole new can of worms.

Also `playlistInfo.GetVideos().FirstOrDefault();` is risky, as the required video may theoretically not be the first in the list or may not even be present at all for some reason (with YouTube anything is possible). It's important to make sure that the ID matches the one the user requested.

Tyrrrz:
Thanks!

PR #448, opened 2020-10-06
Fix httpClient.GetStreamAsync 'to' param out of range issue

MockyJoke:
My service had the following code:
```
segmentedHttpStream.Position = <SomePostionValueWithinRangeOfStreamSize>
await segmentedHttpStream.ReadAsync(buffer, 0, readSize);
```
It throws an error:
```
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'to')
   at System.Net.Http.Headers.RangeItemHeaderValue..ctor(Nullable`1 from, Nullable`1 to)
   at System.Net.Http.Headers.RangeHeaderValue..ctor(Nullable`1 from, Nullable`1 to)
   at YoutubeExplode.ReverseEngineering.YoutubeHttpClient.GetStreamAsync(String requestUri, Nullable`1 from, Nullable`1 to, Boolean ensureSuccess)
   at YoutubeExplode.Internal.SegmentedHttpStream.ReadAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken)
   
```

I'm not sure why the origin code is adding the `Position` to the `_segmentSize `, feels like as long as `Position` is not `0`, it will have a risk of going out of range.
Removing the addition of `Position` in the `to` parameter seem to fix the issue on my side.

### DIFF
--- a/YoutubeExplode/Internal/SegmentedHttpStream.cs
+++ b/YoutubeExplode/Internal/SegmentedHttpStream.cs
@@ -74,7 +74,7 @@ namespace YoutubeExplode.Internal
             // If current stream is not set - resolve it
             if (_currentStream == null)
             {
-                _currentStream = await _httpClient.GetStreamAsync(_url, Position, Position + _segmentSize - 1);
+                _currentStream = await _httpClient.GetStreamAsync(_url, Position, _segmentSize - 1);
             }
 
             // Read from current stream


MockyJoke:
Not sure why Test is failing for videoId:`rsAAeyAr-9Y`. It's strange that the error is on the `from` parameter which I did not change.

I ran the following code on my side and it worked fine, any way to retry the build?
```
StreamManifest streamManifest = await youtubeClient.Videos.Streams.GetManifestAsync("rsAAeyAr-9Y");
IVideoStreamInfo streamInfo = streamManifest.GetMuxed().WithHighestVideoQuality();
await youtubeClient.Videos.Streams.DownloadAsync(streamInfo, "out.mp4");
```

Tyrrrz:
I retried the build and it still fails. Do the tests work locally for you?

> I'm not sure why the origin code is adding the Position to the _segmentSize , feels like as long as Position is not 0, it will have a risk of going out of range.

It has a purpose. If you don't add the position, the `to` value will always be the same, equal to the size of the segment. Which means that if you offset the position past the segment size, you will get corrupt stream.

Tyrrrz:
> It's strange that the error is on the from parameter which I did not change

I think the error message is inaccurate in this case and it refers to the fact that `from` is larger than `to`.

MockyJoke:
> I retried the build and it still fails. Do the tests work locally for you?

Yes, I tried running the follwing code without adding `Position` into `to` and it works locally.
```
StreamManifest streamManifest = await youtubeClient.Videos.Streams.GetManifestAsync("rsAAeyAr-9Y");
IVideoStreamInfo streamInfo = streamManifest.GetMuxed().WithHighestVideoQuality();
await youtubeClient.Videos.Streams.DownloadAsync(streamInfo, "out.mp4");
```

MockyJoke:
I took a closer look into the code and found something interesting:

1. In `StreamClient.GetAsync()`, when the `segmentSize` is set to `long.MaxValue` when the video is not rateLimited
```
var segmentSize = streamInfo.IsRateLimited()
    ? 9_898_989 // this number was carefully devised through research
    : long.MaxValue; // don't use segmentation for non-rate-limited streams

var stream = _httpClient.CreateSegmentedStream(streamInfo.Url, streamInfo.Size.TotalBytes, segmentSize);
```
2. There, the `segmentSize` is being set into the `SegmentedHttpStream` object through constructor.
```
public SegmentedHttpStream(YoutubeHttpClient httpClient, string url, long length, long segmentSize)
{
    _url = url;
    _httpClient = httpClient;
    Length = length;
    _segmentSize = segmentSize;
}
```
3. When `SegmentedHttpStream.ReadAsync()` get called, the current implementation of adding it with `Position` will probably **overflowing** the `_segmentSize` since it's already at `long.MaxValue - 1`
```
public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
{
    // ...
        _currentStream = await _httpClient.GetStreamAsync(_url, Position, Position + _segmentSize - 1);
    // ...
}
```
TL;DR,
It seems like we'll get a long overflow when `_currentStream == null` (probably due to updating the `Position` of the stream, which resets the internal stream)

Tyrrrz:
In this case, it's probably better to actually make segment size nullable. Passing a `long.MaxValue` is a bit of hack.

Btw, how are you using `Position` in your code?

MockyJoke:
> Btw, how are you using `Position` in your code?

I'm using the "Position" of the stream for HTTP range requests.
Let say there's video of with size 123456. The client of my service will be able to make a call like:
```
GET /movie.mp4 HTTP/1.1
Range: bytes=1024-
```
Which my service will set the `Position` of the stream to 1024 before returning the stream to the client, so the client can skip the first 1024 bytes of data.

sample:
```
if(streamFromPosition >= streamInfo.Size.TotalBytes)
{
 // throw exception
}
StreamManifest streamManifest = await youtubeClient.Videos.Streams.GetManifestAsync("rsAAeyAr-9Y");
IVideoStreamInfo streamInfo = streamManifest.GetMuxed().WithHighestVideoQuality();
Stream resultStream = await youtubeClient.Videos.Streams.GetAsync(streamInfo);
resultStream.Position = streamFromPosition ; // streamFromPosition provided by user from the Range header
return new FileStreamResult(resultStream, "video/mp4");
```

Tyrrrz:
Ok makes sense.

If the error you described only happens for non-ratelimited streams, then smartly replacing `long.MaxValue` with `null` should probably do the trick.

MockyJoke:
Accidentally deleted branch. Opening up a new PR at : https://github.com/Tyrrrz/YoutubeExplode/pull/452

PR #451, opened 2020-10-07
Return more specific error message when playlist is unavailable

### LINKED ISSUE #445: Return a more specific error message when playlist is unavailable
Hey. The idea is to get a more detailed error message, of the fail reason from the download. For example, in the case of a private playlists, simply return a response that indicate that. 

Youtube already has this feature: 
image

It would be very nice.
Thank you!

bcwood:
Unfortunately we can't differentiate between the cases of a playlist being private, and one that doesn't exist. In both cases, the response returned is `400 : Bad Request`. This is intended as a fix for #445.

Closes #445


Tyrrrz:
Thanks!

PR #452, opened 2020-10-09
Fix httpClient.GetStreamAsync segmentSize overflow issue

MockyJoke:
Following last comment on PR: https://github.com/Tyrrrz/YoutubeExplode/pull/448
Updated code so `_segmentSize` is nullable. (Will be null for Non-RateLimited video)

### DIFF
--- a/YoutubeExplode/Internal/SegmentedHttpStream.cs
+++ b/YoutubeExplode/Internal/SegmentedHttpStream.cs
@@ -74,16 +74,10 @@ namespace YoutubeExplode.Internal
             // If current stream is not set - resolve it
             if (_currentStream == null)
             {
-                if (_segmentSize == null)
-                {
-                    // The current is not RateLimited - _segmentSize will be null
-                    _currentStream = await _httpClient.GetStreamAsync(_url, Position);
-                }
-                else
-                {
-                    // The current is RateLimited - use _segmentSize with Position to set 'to' stream position
-                    _currentStream = await _httpClient.GetStreamAsync(_url, Position, Position + _segmentSize - 1);
-                }
+                var from = Position;
+                var to = _segmentSize != null ?  Position + _segmentSize - 1 : null;
+                
+                _currentStream = await _httpClient.GetStreamAsync(_url, from, to);
             }
 
             // Read from current stream
--- a/YoutubeExplode/Internal/SegmentedHttpStream.cs
+++ b/YoutubeExplode/Internal/SegmentedHttpStream.cs
@@ -76,7 +76,7 @@ namespace YoutubeExplode.Internal
             {
                 if (_segmentSize == null)
                 {
-                    // The current is NOT RateLimited - _segmentSize will be null
+                    // The current is not RateLimited - _segmentSize will be null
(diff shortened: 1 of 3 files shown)


YoutubeExplode/Internal/SegmentedHttpStream.cs
```diff
@@ -74,7 +74,16 @@ public override async Task<int> ReadAsync(byte[] buffer, int offset, int count,
...
+                if (_segmentSize == null)
+                {
+                    // The current is not RateLimited - _segmentSize will be null
+                    _currentStream = await _httpClient.GetStreamAsync(_url, Position);
+                }
+                else
+                {
+                    // The current is RateLimited - use _segmentSize with Position to set 'to' stream position
+                    _currentStream = await _httpClient.GetStreamAsync(_url, Position, Position + _segmentSize - 1);
+                }
```
Tyrrrz, line 86:
```suggestion
                if (_segmentSize == null)
                {
                    // The current is not RateLimited - _segmentSize will be null
                    _currentStream = await _httpClient.GetStreamAsync(_url, Position);
                }
                else
                {
                    // The current is RateLimited - use _segmentSize with Position to set 'to' stream position
                    _currentStream = await _httpClient.GetStreamAsync(_url, Position, Position + _segmentSize - 1);
                }
```
```suggestion

                var from = Position;
                var to = _segmentSize != null ?  Position + _segmentSize - 1 : null;
                
                _currentStream = await _httpClient.GetStreamAsync(_url, from, to);
```

YoutubeExplode/Videos/Streams/StreamClient.cs
```diff
@@ -284,9 +284,9 @@ public Task<Stream> GetAsync(IStreamInfo streamInfo)
             // so we'll be splitting the stream into small segments and retrieving them separately, to
             // work around rate limiting.
 
-            var segmentSize = streamInfo.IsRateLimited()
+            long? segmentSize = streamInfo.IsRateLimited()
                 ? 9_898_989 // this number was carefully devised through research
-                : long.MaxValue; // don't use segmentation for non-rate-limited streams
+                : default(long?); // don't use segmentation for non-rate-limited streams
```
Tyrrrz, line 289:
```suggestion
            var segmentSize = streamInfo.IsRateLimited()
                ? 9_898_989 // this number was carefully devised through research
                : (long?) null; // don't use segmentation for non-rate-limited streams
```

Tyrrrz:
Added a few comments. Did you check if that fixes your original issue you had?

YoutubeExplode/Videos/Streams/StreamClient.cs
MockyJoke: Agree, fixing it on next commit.

YoutubeExplode/Internal/SegmentedHttpStream.cs
MockyJoke: Agree, fixing it on next commit

MockyJoke:
> Added a few comments. Did you check if that fixes your original issue you had?

Yes, this change fixed the original issue.

Tyrrrz:
Awesome, thanks!

PR #502, opened 2021-02-08
Fix video search and playlist metadata functionality

### LINKED ISSUE #501: Playlist response is broken (search & normal playlists)
Error:
   Playlist response is broken.
   In most cases, this error indicates that the problem is on YouTube's side and this is not a bug in the library.
   To resolve this error, please wait some time and try again.
   If this issue persists, please report it on the project's GitHub page.
YoutubeExplode.Exceptions.TransientFailureException

Code:
`        static void Main(string[] args)
        {
            search();
            Thread.Sleep(10000);
        }
        private static async void search()
        {
            var vids = await new YoutubeClient().Search.GetVideosAsync("random search");
            Console.WriteLine(vids.Count);
        }`

d4n3436:
This PR fixes the problems in the video search and playlist metadata functionality.

After investigating and searching for alternative methods to get videos from a query, I found out that you can get that info scraping directly from YouTube.
When making a search with YouTube and viewing the HTML source, there's an embedded JSON inside a script tag with a lot of useful information about the videos that appears in the results. This also appears when viewing a playlist.
I had to rewrite the Playlist/Search results logic from the lib to extract that JSON and make it to work, it also took me several hours to map every method with its corresponding JSON property.

Almost every data about a video/playlist can be obtained, except the following:
- Like count
- Dislike count
- Keywords
- Upload date (The JSON property returns this info as "x minutes/hours/days/weeks/months/years ago")

These values can probably be obtained in a way that doesn't require scraping another YouTube page.

Since this fix changes the way `SearchClient.GetVideosAsync()` works internally, this makes the parameter `page` useless because there's no "page" to select. I don't know what I can do about this so I left it unchanged.

When running the tests, there are 6 tests that fail because of `PlaylistUnavailableException`. I can't see the playlists in the browser because of the error: "This playlist type is unviewable.". This makes sense because with this change now the lib gets the data scraping the playlist page.

Closes #501.

### DIFF
--- a/YoutubeExplode.Tests/PlaylistSpecs.cs
+++ b/YoutubeExplode.Tests/PlaylistSpecs.cs
@@ -24,14 +24,12 @@ namespace YoutubeExplode.Tests
             playlist.Title.Should().Be("osu! Highlights");
             playlist.Author.Should().Be("Tyrrrz");
             playlist.Description.Should().Be("My best osu! plays");
+            playlist.ViewCount.Should().BeGreaterOrEqualTo(133);
             playlist.Thumbnails?.LowResUrl.Should().NotBeNullOrWhiteSpace();
             playlist.Thumbnails?.MediumResUrl.Should().NotBeNullOrWhiteSpace();
             playlist.Thumbnails?.HighResUrl.Should().NotBeNullOrWhiteSpace();
             playlist.Thumbnails?.StandardResUrl.Should().NotBeNullOrWhiteSpace();
             playlist.Thumbnails?.MaxResUrl.Should().NotBeNullOrWhiteSpace();
-            //playlist.Engagement.ViewCount.Should().BeGreaterOrEqualTo(133);
-            //playlist.Engagement.LikeCount.Should().BeGreaterOrEqualTo(0);
-            //playlist.Engagement.DislikeCount.Should().BeGreaterOrEqualTo(0);
         }
 
         [Fact]
--- a/YoutubeExplode.Tests/PlaylistSpecs.cs
+++ b/YoutubeExplode.Tests/PlaylistSpecs.cs
@@ -29,9 +29,9 @@ namespace YoutubeExplode.Tests
             playlist.Thumbnails?.HighResUrl.Should().NotBeNullOrWhiteSpace();
             playlist.Thumbnails?.StandardResUrl.Should().NotBeNullOrWhiteSpace();
             playlist.Thumbnails?.MaxResUrl.Should().NotBeNullOrWhiteSpace();
-            playlist.Engagement.ViewCount.Should().BeGreaterOrEqualTo(133);
-            playlist.Engagement.LikeCount.Should().BeGreaterOrEqualTo(0);
-            playlist.Engagement.DislikeCount.Should().BeGreaterOrEqualTo(0);
+            //playlist.Engagement.ViewCount.Should().BeGreaterOrEqualTo(133);
+            //playlist.Engagement.LikeCount.Should().BeGreaterOrEqualTo(0);
+            //playlist.Engagement.DislikeCount.Should().BeGreaterOrEqualTo(0);
         }
 
         [Fact]
@@ -59,7 +59,7 @@ namespace YoutubeExplode.Tests
         [Theory]
         [InlineData("PLI5YfMzCfRtZ8eV576YoY3vIYrHjyVm_e")] // normal
         [InlineData("RD1hu8-y6fKg0")] // video mix
-        [InlineData("RDMMU-ty-2B02VY")] // my mix
+        //[InlineData("RDMMU-ty-2B02VY")] // my mix
         [InlineData("RDCLAK5uy_lf8okgl2ygD075nhnJVjlfhwp8NsUgEbs")] // music mix
         [InlineData("OLAK5uy_lLeonUugocG5J0EUAEDmbskX4emejKwcM")] // music album
         [InlineData("PL601B2E69B03FAB9D")]
@@ -103,12 +103,12 @@ namespace YoutubeExplode.Tests
         [InlineData("PLI5YfMzCfRtZ8eV576YoY3vIYrHjyVm_e")] // normal
         [InlineData("PLWwAypAcFRgKFlxtLbn_u14zddtDJj3mk")] // large
         [InlineData("OLAK5uy_mtOdjCW76nDvf5yOzgcAVMYpJ5gcW5uKU")] // large 2
-        [InlineData("RD1hu8-y6fKg0")] // video mix
-        [InlineData("RDMMU-ty-2B02VY")] // my mix
+        //[InlineData("RD1hu8-y6fKg0")] // video mix
+        //[InlineData("RDMMU-ty-2B02VY")] // my mix
         [InlineData("RDCLAK5uy_lf8okgl2ygD075nhnJVjlfhwp8NsUgEbs")] // music mix
-        [InlineData("ULl6WWX-BgIiE")] // channel video mix
+        //[InlineData("ULl6WWX-BgIiE")] // channel video mix
         [InlineData("UUTMt7iMWa7jy0fNXIktwyLA")] // user uploads
-        [InlineData("PUTMt7iMWa7jy0fNXIktwyLA")] // popular user uploads
+        //[InlineData("PUTMt7iMWa7jy0fNXIktwyLA")] // popular user uploads
         [InlineData("OLAK5uy_lLeonUugocG5J0EUAEDmbskX4emejKwcM")] // music album
         [InlineData("PL601B2E69B03FAB9D")]
         public async Task I_can_get_videos_included_in_any_available_YouTube_playlist(string playlistId)
--- a/YoutubeExplode.Tests/PlaylistSpecs.cs
+++ b/YoutubeExplode.Tests/PlaylistSpecs.cs
@@ -58,8 +58,6 @@ namespace YoutubeExplode.Tests
 
         [Theory]
         [InlineData("PLI5YfMzCfRtZ8eV576YoY3vIYrHjyVm_e")] // normal
-        [InlineData("RD1hu8-y6fKg0")] // video mix
-        //[InlineData("RDMMU-ty-2B02VY")] // my mix
         [InlineData("RDCLAK5uy_lf8okgl2ygD075nhnJVjlfhwp8NsUgEbs")] // music mix
         [InlineData("OLAK5uy_lLeonUugocG5J0EUAEDmbskX4emejKwcM")] // music album
         [InlineData("PL601B2E69B03FAB9D")]
@@ -103,12 +101,8 @@ namespace YoutubeExplode.Tests
         [InlineData("PLI5YfMzCfRtZ8eV576YoY3vIYrHjyVm_e")] // normal
         [InlineData("PLWwAypAcFRgKFlxtLbn_u14zddtDJj3mk")] // large
         [InlineData("OLAK5uy_mtOdjCW76nDvf5yOzgcAVMYpJ5gcW5uKU")] // large 2
-        //[InlineData("RD1hu8-y6fKg0")] // video mix
-        //[InlineData("RDMMU-ty-2B02VY")] // my mix
         [InlineData("RDCLAK5uy_lf8okgl2ygD075nhnJVjlfhwp8NsUgEbs")] // music mix
-        //[InlineData("ULl6WWX-BgIiE")] // channel video mix
         [InlineData("UUTMt7iMWa7jy0fNXIktwyLA")] // user uploads
-        //[InlineData("PUTMt7iMWa7jy0fNXIktwyLA")] // popular user uploads
         [InlineData("OLAK5uy_lLeonUugocG5J0EUAEDmbskX4emejKwcM")] // music album
         [InlineData("PL601B2E69B03FAB9D")]
         public async Task I_can_get_videos_included_in_any_available_YouTube_playlist(string playlistId)
--- a/YoutubeExplode/AccessibilityExtensions.cs
+++ b/YoutubeExplode/AccessibilityExtensions.cs
@@ -2,6 +2,7 @@ using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using YoutubeExplode.Internal.Extensions;
+using YoutubeExplode.Playlists;
 using YoutubeExplode.Videos;
 
 namespace YoutubeExplode
@@ -18,15 +19,33 @@ namespace YoutubeExplode
             await asyncVideoEnumerable.ToListAsync();
 
         /// <summary>
+        /// Buffers the asynchronous list of playlist videos in memory.
+        /// </summary>
+        public static async ValueTask<IReadOnlyList<PlaylistVideo>> BufferAsync(this IAsyncEnumerable<PlaylistVideo> asyncVideoEnumerable) =>
+            await asyncVideoEnumerable.ToListAsync();
+
+        /// <summary>
         /// Buffers the asynchronous list of videos in memory, up to the specified number of videos.
         /// </summary>
         public static async ValueTask<IReadOnlyList<Video>> BufferAsync(this IAsyncEnumerable<Video> asyncVideoEnumerable, int count) =>
             await asyncVideoEnumerable.TakeAsync(count).BufferAsync();
 
         /// <summary>
+        /// Buffers the asynchronous list of playlist videos in memory, up to the specified number of videos.
+        /// </summary>
+        public static async ValueTask<IReadOnlyList<PlaylistVideo>> BufferAsync(this IAsyncEnumerable<PlaylistVideo> asyncVideoEnumerable, int count) =>
+            await asyncVideoEnumerable.TakeAsync(count).BufferAsync();
+
+        /// <summary>
         /// Gets the awaiter that encapsulates an operation that buffers a list of videos in-memory,
         /// </summary>
         public static ValueTaskAwaiter<IReadOnlyList<Video>> GetAwaiter(this IAsyncEnumerable<Video> asyncVideoEnumerable) =>
+            asyncVideoEnumerable.BufferAsync().GetAwaiter();
+
+        /// <summary>
+        /// Gets the awaiter that encapsulates an operation that buffers a list of playlist videos in-memory,
+        /// </summary>
+        public static ValueTaskAwaiter<IReadOnlyList<PlaylistVideo>> GetAwaiter(this IAsyncEnumerable<PlaylistVideo> asyncVideoEnumerable) =>
             asyncVideoEnumerable.BufferAsync().GetAwaiter();
     }
 }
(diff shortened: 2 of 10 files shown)


Tyrrrz:
Hey, thanks for taking a stab at this.
Sadly, we do need those missing properties as well to stay backwards compatible.

Thought maybe mobile search results are easier to parse, but it seems they return the same data:
https://m.youtube.com/results?search_query=hello+world

d4n3436:
> Thought maybe mobile search results are easier to parse, but it seems they return the same data

Yes, I checked the embedded JSON using this [url](https://m.youtube.com/results?search_query=hello+world&hl=en&persist_hl=1&persist_app=1&app=m) (it forces English language and mobile version) and both return the same data with a very similar structure.

The only way to get the missing data would be using `VideoClient.GetAsync()`.

Maybe adding a warning message about the properties not having a value when using `SearchClient.GetVideosAsync()` and recommend using `VideoClient.GetAsync()` to get the missing data would be reasonable?

Another option would be using that method inside `SearchClient.GetVideosAsync()`.

Tyrrrz:
Calling `VideoClient.GetAsync()` implicitly is definitely bad because it's slow. Let the user decide whether to call it or not.
It seems that we'd have to drop the missing data and move the videos to a new class (`PlaylistVideo` or maybe a better name that also includes search results?). Playlists are also broken (some are still working, but probably not for long; I suspect YT is rolling the update in waves), so we'll probably use the same approach there as well.

d4n3436:
I moved the resulting videos from playlists and search results to a new class (`PlaylistVideo`) and moved the `ViewCount` property out of the `Engagement` property in `PlaylistVideo` and `Playlist` because it wouldn't make sense to have a property with only 1 of 3 available values.

I believe the PR is ready for review/merge, although there may be some room for improvement regarding a way to get multiple pages from the search results, if there's any.

CCRed95:
Hey @d4n3436, I used your pull request code to pack an experimental/pre-release nuget package to get my youtube downloader app working.

One thing i did is I added an interface for Video and PlaylistVideo of the common properties between them, and I noticed that `DateTimeOffset UploadDate { get; }` is not available in `PlaylistVideo`, i tried to add it but I am not familiar enough with the internals of this library to figure it out...

But a more important issue: When I'm calling `.GetVideosAsync(Channeld)`, sometimes not all of videos will be returned.  

If I try to load up this channel: https://www.youtube.com/channel/UCr6mdNMsNBaMwVczVujRuoQ
I get only 96 videos, and on my friend's computer he got 100 videos, but that channel actually has 976 videos. I've attached my downloader app with the channel loaded up:

image

> I believe the PR is ready for review/merge, although there may be some room for improvement regarding a way to get multiple pages from the search results, if there's any.

Is this related to getting videos from a channel as well? Maybe it's loading only 1 page properly and it caps it at 100 (or 96 videos)?

d4n3436:
> One thing i did is I added an interface for Video and PlaylistVideo of the common properties between them, and I noticed that `DateTimeOffset UploadDate { get; }` is not available in ` PlaylistVideo` , i tried to add it but I am not familiar enough with the internals of this library to figure it out...

This is because the new method of obtaining the data does not provide that info.

> Is this related to getting videos from a channel as well? Maybe it's loading only 1 page properly and it caps it at 100 (or 96 videos)?

I believe the limit of 100 videos affects search results, playlists and channel videos.
There seems to be a way to bypass that limit based on what I've seen in other YouTube scraping libraries.
I will try to implement that bypass but I will have to do more testing and changes.

Tyrrrz:
Hmm, how is pagination handled in your approach @d4n3436?

d4n3436:
@Tyrrrz I didn't write the code yet, but the process should be straighforward for playlists:

- Make a first request to the playlist page to get the first 100 videos and the continuation token.
- Use the token to get the next videos from InnerTube.
- Repeat the 2nd step until there are no videos left.

 For search results it might be harder because there are parameters to specifiy a starting page and a page count, and the videos are returned in batches of 100.

CCRed95:
@d4n3436 I am stepping through the code for trying to understand how it works and why it's not able to paginate, and I think I'm seeing the issue. Let me know if I'm wrong in my understanding of this, I am new to this source code!

My understanding: if you query `ChannelClient.GetUploadsAsync(ChannelId)`, it really defers that to the `PlaylistClient`. It takes the Channel ID, chops off the beginning `"UC"` from the channel ID and replacing it with a `"UU"`, which looks like it returns just a playlist that contains ALL the videos that the channel uploaded. In `.GetVideosAsync(PlaylistId)`, you have a while loop that is supposed to continue querying for the `PlaylistResponse` but with an `index` offset...

The issue is, that you're tracking new encountered video ID's with the hashset, and increasing that `index` by the `countdelta`, which it all adds up logically... but what actually happens is it queries for the first chunk of 100, adds +100 to `index`, and then the 2nd time it queries, even though you're passing in `index` = 100 as the pagination offset, YouTube still returns the exact same thing, it doesn't list videos in the playlist starting at that provided index.

Even just in the web browser I can see that the `index=100` url argument does not affect what YouTube returns at all.

(Url's used:)
https://www.youtube.com/playlist?list=UUlha3awlu0QFQzlS-9oi_Fg&index=0&hl=en&persist_hl=1
https://www.youtube.com/playlist?list=UUlha3awlu0QFQzlS-9oi_Fg&index=100&hl=en&persist_hl=1

image

On the left side, I have it querying for `index=0`. on the right side, I have it querying for `index=100`. You can see that the results list is not offset by `100`, they're identical. There are 3,120 videos in this playlist, so it should be looping ~32 times before it's done loading, but it only loops twice, once getting the first 100, the 2nd time getting the SAME 100 videos, and since you're checking for identical ID's in the `HashSet`, it sees all of those videos as already processed, and the `countDelta` for that batch is `0`, so it thinks that it's done and breaks the while.

Did the `index` url argument being set to `100` or whatever used to work, and now YouTube has changed the way that page works possibly?

Thanks!! Appreciate your guys' help.
-CCRed95

d4n3436:
> Did the `index` url argument being set to `100` or whatever used to work, and now YouTube has changed the way that page works possibly?

@CCRed95 Yes, the `index` argument was used in the API that YouTube broke [here](https://github.com/Tyrrrz/YoutubeExplode/blob/807668391de8dbd5f3da18d506610d6bb089b467/YoutubeExplode/ReverseEngineering/Responses/PlaylistResponse.cs#L129). I left it there because i thought it would work with the playlist page but unfortunately it didn't.

CCRed95:
@d4n3436 Yeah, i've been digging in to YouTube documentation trying to find ANY information about  them deciding to remove the `index` argument from the `youtube.com/playlist?` page, you would think that if YouTube was going to remove the ability to set the playlist index, or changed how you're supposed to do it or renamed the argument, there would be some mention of it in their documentation online or someone would have mentioned that it's no longer working, I did a bunch of Google searches and filtered by "Last Month" and I don't see anything really. 

I do see people back prior to the change on a bunch of StackOverflow posts using the `index` parameter exactly how it's being used here, to for example set the current video of a playlist to a random index. So presumably all of heir code for that is also broken by this change?

I can't even really find a very good document from YouTube or Google with documentation of what arguments are available for each page (playlist, watch, search, embed, etc.), I only see like `iframe` embed source documentation which I couldn't get to work either with:

https://www.youtube.com/embed/videoseries?list=UUlha3awlu0QFQzlS-9oi_Fg&index=100

or various other variants I tried. Also the `embed` for a `videoseries` playlist seems to have the playlist load, but it's limited at a fixed `200` videos and it doesn't even have `continuationTrigger` type stuff where when you scroll to the bottom of the list, it fetches another 200 like how `playlist` pages work.

d4n3436:
@CCRed95 It took me a while, but I was able to implement the "bypass" (actually the way YouTube gets the videos in their page) and it can get all the 3120 videos from the playlist.
I will continue working on the next part (search results) tomorrow.

image

d4n3436:
@CCRed95 With the last commit you should be able to get all the videos in the playlist.

@Tyrrrz Is this the correct way to paginate the videos in [SearchClient.GetVideosAsync()](https://github.com/d4n3436/YoutubeExplode/blob/playlist_fix/YoutubeExplode/Search/SearchClient.cs)?
The internal API returns the videos in batches of 100 (about 20 for search results), and it doesn't stop until there are no videos left (the continuation token is missing).

About the API key, I think its safe to hard-code it, since YouTube is using the same key for years.

CCRed95:
@d4n3436  Bro! Thank you very much, you are the man. Really appreciate you putting in the work this fast.

We archive comedy radio and podcasts shows, apparently a lot of the guys that have huge channels of this content are getting a lot of heat from YouTube, really unreasonable community guidelines and copyright strikes from YouTube, DMCA's from SiriusXM and other media companies in the last like few weeks, when this content has been up for YEARS without any problem...

Like 3 or 4 of the big channels got notified that they're all getting banned effective Feb. 25, so we're scrambling to get all this content saved locally so we don't lose it forever! So I really appreciate you jumping on this fast, seems to grab all videos and matches up to their upload count perfectly. My app works great now!

image

CCRed95:
@d4n3436 A few comments about your PR and the library now though...

You mentioned in your #502 issue this:
>Almost every data about a video/playlist can be obtained, except the following:
>- Like count
>- Dislike count
>- Keywords
>- Upload date (The JSON property returns this info as "x minutes/hours/days/weeks/months/years ago")

My app did rely on the `UploadDate` because I have some ID3 .mp3 metadata formatting, custom file naming conventions and subfolder structuring by year/month, etc., so I had to take that functionality out at least temporarily. 

I tried circumventing this by doing a `var video = client.Videos.GetAsync(videoId)` call and grabbing the `video.UploadDate`, but it obviously had an obscenely high performance cost on channels that have thousands of videos, and in some cases it would end up just hanging on that call or exceeding rate limiter.

>These values can probably be obtained in a way that doesn't require scraping another YouTube page.

I ended up seeing this only after spending a bunch of time screwing around with trying different ways of getting this data without doing a full `watch?` page scrape / GetAsync(...) call just for that, and I haven't had much success. (I'm still learning the guts of YouTube and their APIs and trying to get a good grasp on the ReverseEngineering / httpclient scraping in this library.)

Do you have any pointers on what direction I should look in for how this would be implemented in an efficient way? It'd be nice to not have to handle 2 video metadata class types, or at least making a shared `IVideoMetadata` interface that contains as many properties as possible, or even just shared properties + the `UploadDate`!

Thanks again, man.
--CC

d4n3436:
@CCRed95 Unfortunately the only known way to get the missing metadata is using `VideoClient.GetAsync()`.

> Do you have any pointers on what direction I should look in for how this would be implemented in an efficient way?

There's a way to speed up the process, but it implies modifying the source code.
First, go the file `VideoClient.cs` in `/YoutubeExplode/Videos`
The method `GetVideoFromWatchPageAsync()` calls 2 other methods to get the metadata, the first method uses a YouTube API endpoint and the second method scrapes the watch page.
Removing the call to the watch page (`var watchPage = await WatchPage.GetAsync(_httpClient, id)`) should improve the performance:

```cs
private async Task<Video> GetVideoFromWatchPageAsync(VideoId id)
{
    var videoInfoResponse = await VideoInfoResponse.GetAsync(_httpClient, id);
    var playerResponse = videoInfoResponse.GetPlayerResponse();

    return new Video(
        id,
        playerResponse.GetVideoTitle(),
        playerResponse.GetVideoAuthor(),
        playerResponse.GetVideoChannelId(),
        playerResponse.GetVideoUploadDate(),
        playerResponse.GetVideoDescription(),
        playerResponse.GetVideoDuration(),
        new ThumbnailSet(id),
        playerResponse.GetVideoKeywords(),
        new Engagement(0, 0, 0)
    );
}
```
You can also add a new method that only returns the upload date.

> It'd be nice to not have to handle 2 video metadata class types, or at least making a shared `IVideoMetadata` interface that contains as many properties as possible, or even just shared properties + the UploadDate!

I like this idea, having a common interface like `IVideo`, which `Video` and `PlaylistVideo` derives from.

Tyrrrz:
> @CCRed95 With the latest commit you should be able to get all the videos in the playlist.
> 
> @Tyrrrz Is this the correct way to paginate the videos in [SearchClient.GetVideosAsync()](https://github.com/d4n3436/YoutubeExplode/blob/playlist_fix/YoutubeExplode/Search/SearchClient.cs)?
> The internal API returns the videos in batches of 100 (about 20 for search results), and it doesn't stop until there are no videos left (the continuation token is missing).
> 
> About the API key, I think it's safe to hard-code it, since YouTube is using the same key for years.

Nice!
I think at this point we can remove `int startPage` parameter as there's no reliable way to start from a specific page, if I understand correctly.

> I  ended up seeing this only after spending a bunch of time screwing around with trying different ways of getting this data without doing a full watch? page scrape / GetAsync(...) call just for that, and I haven't had much success. (I'm still learning the guts of YouTube and their APIs and trying to get a good grasp on the ReverseEngineering / httpclient scraping in this library.)

@CCRed95 essentially, if the date is not shown in the playlist anywhere on YouTube, it's probably not obtainable. As far as I can tell, it only shows it in a form of "A month ago" or similar, which is too inaccurate to even bother parsing.

The way things are going, it looks like `PlaylistVideo` will have a rather limited subset of the data available in `Video`, unfortunately.

Tyrrrz:
@d4n3436 thanks for your help!
Do you think your work in this PR is done or is there something else you want to do?
I'm planning to merge it, release a prerelease version to unblock people, and then take some time to refactor the whole library (it's been a long time coming, and with the recent changes it can't be delayed much longer).

d4n3436:
@Tyrrrz I think the PR is ready to be merged, I did a lot of tests and I'm using my build with my Discord bot for a week and I didn't find any error except one minor bug in [GetDuration()](https://github.com/d4n3436/YoutubeExplode/blob/30e58e7f8b0c6161bd20c3b23e896f7025fa3061/YoutubeExplode/ReverseEngineering/Responses/PlaylistResponse.cs#L181) that occurs with videos longer than 24 hours.
I modified this method so that it returns the default value instead of throwing an exception because I don't know how to handle that case.

Tyrrrz:
@d4n3436 thanks!
How does the duration look in `simpleText` when it's above 24hours?

d4n3436:
> How does the duration look in `simpleText` when it's above 24hours?

@Tyrrrz it looks just like any other duration (like `30:12:23`). There are also videos longer that 100 hours (`200:00:00`).
Here are [some examples](https://www.youtube.com/results?search_query=longest+video).

Tyrrrz:
> > How does the duration look in `simpleText` when it's above 24hours?
> 
> @Tyrrrz it looks just like any other duration (like `30:12:23`). There are also videos longer that 100 hours (`200:00:00`).
> Here are [some examples](https://www.youtube.com/results?search_query=longest+video).

Do you have any idea why the parsing failed in those cases?

d4n3436:
I guess it fails because `TimeSpan.ParseExact` does not allow the time format (`hh:mm:ss`) to have more than 24 hours in the `hh` part.
One way to fix this might be converting the hours to days.

CCRed95:
@Tyrrrz 
>The way things are going, it looks like `PlaylistVideo` will have a rather limited subset of the data available in `Video`, unfortunately.

Aww, that is too bad.. I "Fixed this" by doing a single playlist / channel query for all `IEnumerable<(videoID, DateTimeOffset)>`'s from the youtube v3 API directly (obviously not a permanent or even share-able fix) and wrote a matching system, so that 1 query will take care of a large amount of `DateTimeOffset UploadTime { get; }` 's, which is ugly, but It's all I could think of right now.

Thanks to you both for all your hard work on this project. <3

CCRed95:
@Tyrrrz are we having a problem with the `Duration` parsing on short `Durations` with special formatting? Couldn't we write different format recognizers for the different syntaxes, and have them determine the syntax, and parse _based_ on client's local time (or global UTC time)? 

- Is it that `3 hours ago` (`DateTime.UtcNow` - `timeAgo`) = `approxUploadDate` is not accurate / reliable
   enough a `DateTime` to show?
- But `30:12:23` `'Durations`, the older durations, are accurate?

Thanks for the help and putting up with my noob basic questions!!

Tyrrrz:
@CCRed95 from my understanding, the relative dates lose accuracy as they go further in the past. It starts with X hours ago which is accurate to the hour, but then, when it gets to X months or X years ago, the accuracy drops to month or year (rounded up even) respectively. In my opinion, that information is way too inaccurate to be useful in any way.

CCRed95:
@Tyrrrz Ahhh. that's right. Goofy ass youtube making life difficult again :)

JaykeBird:
Thank you for all the work with this! As was discussed previously, it definitely would be helpful to have some wrapper class or interface that Video and PlaylistVideo both take from, with access to at least as many properties as the two classes share. I can probably go in and work and make a PR for this myself later this week.

CCRed95:
@JaykeBird I implemented this locally, and it seems to work well. I also modified a few other minor things that seem to me like they would make sense. 
If I get to it before you, I could also create this in a PR sometime this weekend, since it would be nice to have those two extremely similar types share an interface/base class, (especially for how I am using the library), but also just seems like a logically sound minor structural change.

maerf:
@CCRed95 So funny to see somebody posting here also trying to download O&A videos. I really laughted when I saw O&A in your screenshots ... that's what youtube is for. I'm trying to download all Louie/Bill Burr/Patrice/Stalker Patti appearances.
I didn't know there is an immediate threat of takedown currently.  
Is your "YoutubeRipper / Metadata-Miner" available somewhere? Google didn't help.
Playlist download has been broken for weeks as far as I can see.
@Tyrrrz mentioned an alpha version , but I don't know how to get it (and if it fixed the playlist problem).

Tyrrrz:
@maerf this should have instructions how to install it:
https://www.nuget.org/packages/YoutubeExplode/6.0.0-alpha

CCRed95:
@maerf Lol, pleasure to see another O&A guy in these parts! YES. Yes. Yes, there is an immediate threat to O&A content on YouTube right now. I run a Discord server with a lot of the big name channels on YouTube who have been archiving O&A / R&F / etc. content for years, decades even. Big channels are being targeted and we've already had 2 guys go down, one from targeted DMCA's by SiriusXM, the other from "Hate Speech" or ToS violations. You should join our discord group of archivists, we'd love to have you! 

I have a full O&A / R&F / Cumtown / Ricky Gervais Show / Tough Crowd with Colin Quinn archive hosted on archive.org. I am currently writing a web app / service that's called podcache.io (used to be called opieandanthonylive.com, which was active for about a year and a half before GoDaddy couldn't support my bandwidth needs), that is basically "The Spotify for Comedy Radio, Podcasts, and Rare TV" which will include literally every episode of:

- Opie and Anthony
- Ron and Fez
- Cumtown
- Tough Crowd with Colin Quinn
- Ricky Gervais Show (XFM, Podcast Series, and HBO Show)
- Matt and Shane Secret Podcast
- Than and Sam Show
- Unmasked
- Bob Kelly and Colin Quinn Podcast
- Saturday Night Virus
- And much more...

They'll be available for Streaming, Downloading as .mp3/.mp4 in bulk, dynamic generation of playlists, full integration with show rundowns, guest tags, bit tags, account system for tracking what you've listened to and will pick up where you left off in playback (like Netflix or YouTube does), commenting, favoriting, tons more. And the service will be totally free!

image

maerf:
It looks great. Sounds almost to good to be true lol. How can I join the discord?

CCRed95:
@maerf Awesome!! Direct Message me on:

-   Twitter: https://twitter.com/CCRed95 
-   Reddit: https://www.reddit.com/u/CCRed95/
-   Or just add me as a friend on Discord directly: **CCRed95#6710**

And I'll send you over an invite link!

PR #517, opened 2021-03-29
Don't escape search query

Baimbekka:
Youtube does not escape the query for the search call.  The added test fails without this change.


Tyrrrz:
Yeah, this looks like it was left over from the time when the query was passed as URL parameter. Thanks!

Can you please remove unrelated files from the PR and revert the formatting change which replaced spaces with tabs?

Baimbekka:
Done

Tyrrrz:
Thanks :)

PR #607, opened 2021-12-08
Ignore braces inside strings in `Json.Extract`

wleader:
Makes Json.Extract aware of whether or not the current character is inside of a string, and only count braces a that are not inside of sting.

This deals specifically with videos that have unmatched braces inside strings in their data. Example: watch?v=zSgiXGELjbc


Tyrrrz:
Cool! Thanks.

PR #612, opened 2022-01-13
Fix mix playlists

### LINKED ISSUE #555: Metadata can't be resolved for mix playlists
### Version

v6.0.2

### Details

When trying to get videos from a playlist I get a PlaylistUnavailableException in `public async ValueTask<PlaylistExtractor> GetPlaylistAsync(...)` [ YoutubeController.cs]
Examples:
https://www.youtube.com/watch?v=kKDO8Rxw3vs&list=RDkKDO8Rxw3vs&start_radio=1 -> Playlist 'RDkKDO8Rxw3vs' is not available.
https://www.youtube.com/watch?v=3U4lQ8xX3ag&list=RDEMPeuXuiddECzwmszjRj1EtQ  -> Playlist 'RDEMPeuXuiddECzwmszjRj1EtQ' is not available.

### Steps to reproduce

Insert one of the above mentioned URLs in Tyrrz/YoutubeDownloader.

xBaank:
Closes #555

Fix issue where metadata couldn't be resolved for mix playlists.

- Added ```https://www.youtube.com/youtubei/v1/next``` which seems to be the only endpoint to resolve mix playlits, as it can resolve any type of playlist  I changed the way playlistExtractor and playlistClient works. 
- Replaced  continuation token for visitorData (this is because continuationToken can't be used with Mix playlists).
- Added support to get thumbnails with system playlists (System playlists couldn't resolve thumbnails metadata). In the case of mix playlists it uses the first video thumbnail as youtube does too.
- Playlists metadata is tried to be resolved with the browse endpoint, if it can't resolve it (case of mix playlists) it will try to resolve it with next endpoint.
- Playlists url is now passed as a parameter to the constructor only in case that the browse endpoint can't get the playlists and gets it from next endpoint.


YoutubeExplode.Tests/TestData/PlaylistIds.cs
```diff
@@ -11,4 +11,5 @@ internal static class PlaylistIds
     public const string UserUploads = "UUTMt7iMWa7jy0fNXIktwyLA";
     public const string Weird = "PL601B2E69B03FAB9D";
     public const string ContainsLongVideos = "PLkk2FsMngwGi9FNkWIoNZlfqglcldj_Zs";
+    public const string MixPlaylist = "RDTsYhxMnGYCw";
```
Tyrrrz, line 14: Can you please rename this to `VideoMix` and move it above `MusicMix`? Just to keep things consistent. All playlists starting with `RD` are mix playlists, but they behave a bit differently.

YoutubeExplode.DemoConsole/Program.cs
```diff
@@ -21,6 +21,7 @@ public static async Task<int> Main()
         Console.Write("Enter YouTube video ID or URL: ");
         var videoId = VideoId.Parse(Console.ReadLine() ?? "");
 
+
```
Tyrrrz, line 24:
```suggestion

```

Remove extra linebreak

YoutubeExplode/Bridge/PlaylistExtractor.cs
```diff
@@ -20,6 +20,20 @@ internal partial class PlaylistExtractor
...
+            .GetPropertyOrNull("twoColumnWatchNextResults")?
+            .GetPropertyOrNull("playlist")?
+            .GetPropertyOrNull("playlist") ??
+
+        _content
+            .GetPropertyOrNull("continuationContents")?
+            .GetPropertyOrNull("playlistPanelContinuation")
+
+
+    );
```
Tyrrrz, line 35:
```suggestion
            .GetPropertyOrNull("playlistPanelContinuation")
    );
```

Remove extra linebreak

YoutubeExplode/Bridge/PlaylistExtractor.cs
```diff
@@ -33,17 +47,19 @@ internal partial class PlaylistExtractor
             .ElementAtOrNull(1)?
             .GetPropertyOrNull("playlistSidebarSecondaryInfoRenderer")
     );
-
-    public bool IsPlaylistAvailable() => Memo.Cache(this, () =>
+    public bool IsPlaylistVideosAvailable() => Memo.Cache(this, () =>
+        TryGetPlaylistProperty() is not null
```
Tyrrrz, line 51:
```suggestion

    public bool IsPlaylistVideosAvailable() => Memo.Cache(this, () =>
        TryGetPlaylistProperty() is not null
```

Add linebreak

YoutubeExplode/Bridge/PlaylistExtractor.cs
```diff
@@ -33,17 +47,19 @@ internal partial class PlaylistExtractor
...
+    );
+    public bool IsPlaylistDetailsAvailable() => Memo.Cache(this, () =>
         TryGetSidebar() is not null
```
Tyrrrz, line 54:
```suggestion

    public bool IsPlaylistDetailsAvailable() => Memo.Cache(this, () =>
        TryGetSidebar() is not null
```

Add linebreak

YoutubeExplode/Bridge/PlaylistExtractor.cs
```diff
@@ -33,17 +47,19 @@ internal partial class PlaylistExtractor
...
 
     public string? TryGetPlaylistTitle() => Memo.Cache(this, () =>
-        TryGetSidebarPrimary()?
+        // next endpoint part
+        TryGetPlaylistProperty()?
             .GetPropertyOrNull("title")?
-            .GetPropertyOrNull("simpleText")?
             .GetStringOrNull() ??
-
+        // browse endpoint part
```
Tyrrrz, line 62: Should we still have the `title.simpleText` as fallback?

YoutubeExplode/Bridge/PlaylistExtractor.cs
```diff
@@ -57,9 +73,17 @@ internal partial class PlaylistExtractor
...
+
     );
 
     public string? TryGetPlaylistAuthor() => Memo.Cache(this, () =>
+        //next enpoint part
+        TryGetPlaylistProperty()?
+            .GetPropertyOrNull("ownerName")?
+            .GetPropertyOrNull("simpleText")?
+            .GetStringOrNull() ??
+        //browse endpoint part
```
Tyrrrz, line 86:
```suggestion
            .GetStringOrNull() ??

        //browse endpoint part
```

Linebreak

YoutubeExplode/Bridge/PlaylistExtractor.cs
```diff
@@ -106,60 +131,80 @@ internal partial class PlaylistExtractor
...
+                .GetPropertyOrNull("thumbnails")?
+                .EnumerateArrayOrNull()?
+                .Select(j => new ThumbnailExtractor(j))
+                .ToArray() ??
+
+        //next endpoint part
+        GetVideos()
+            .FirstOrDefault()?
+            .GetVideoThumbnails()
+            .ToArray() ??
```
Tyrrrz, line 148: Should the `next` part be first, consistent with the others?

YoutubeExplode/Bridge/PlaylistExtractor.cs
```diff
@@ -106,60 +131,80 @@ internal partial class PlaylistExtractor
...
-        // Continuation response
-        _content
-            .GetPropertyOrNull("onResponseReceivedActions")?
-            .EnumerateArrayOrNull()?
-            .ElementAtOrNull(0)?
-            .GetPropertyOrNull("appendContinuationItemsAction")?
-            .GetPropertyOrNull("continuationItems")
+    public string? TryGetPlayListUrl() => Memo.Cache(this, () =>
+     TryGetPlaylistProperty()?
+         .GetPropertyOrNull("playlistShareUrl")?
```
Tyrrrz, line 155: Does this work for both old and new approaches?

YoutubeExplode/Bridge/PlaylistExtractor.cs
```diff
@@ -106,60 +131,80 @@ internal partial class PlaylistExtractor
...
+         .GetStringOrNull()
     );
 
     public IReadOnlyList<PlaylistVideoExtractor> GetVideos() => Memo.Cache(this, () =>
-        TryGetContentRoot()?
+        TryGetPlaylistProperty()?
+            .GetPropertyOrNull("contents")?
             .EnumerateArrayOrNull()?
-            .Select(j => j.GetPropertyOrNull("playlistVideoRenderer"))
+            .Select(j => j.GetPropertyOrNull("playlistPanelVideoRenderer"))
```
Tyrrrz, line 163: Does this work for both old and new approaches?

YoutubeExplode/Bridge/PlaylistExtractor.cs
```diff
@@ -106,60 +131,80 @@ internal partial class PlaylistExtractor
...
 
+    public string? TryGetLastVideoId() => Memo.Cache(this, () =>
+        TryGetPlaylistProperty()?
+            .GetPropertyOrNull("contents")?
+            .EnumerateArrayOrNull()?
+            .Select(j => j.GetPropertyOrNull("playlistPanelVideoRenderer"))
+            .LastOrDefault()?
+            .GetPropertyOrNull("videoId")?
+            .GetStringOrNull()
+    );
```
Tyrrrz, line 179: Can we just use `GetVideos()` here?

YoutubeExplode/Channels/ChannelClient.cs
```diff
@@ -83,6 +83,6 @@ private Channel Extract(ChannelPageExtractor channelPage)
     {
         // Replace 'UC' in channel ID with 'UU'
         var playlistId = "UU" + channelId.Value.Substring(2);
-        return new PlaylistClient(_http).GetVideosAsync(playlistId, cancellationToken);
+        return new PlaylistClient(_http).GetVideosAsync(playlistId, cancellationToken : cancellationToken);
```
Tyrrrz, line 86: Was this change necessary?

```suggestion
        return new PlaylistClient(_http).GetVideosAsync(playlistId, cancellationToken);
```

YoutubeExplode/Playlists/Playlist.cs
```diff
@@ -13,7 +13,7 @@ public class Playlist : IPlaylist
     public PlaylistId Id { get; }
 
     /// <inheritdoc />
-    public string Url => $"https://www.youtube.com/playlist?list={Id}";
+    public string Url => _url ?? $"https://www.youtube.com/playlist?list={Id}";
```
Tyrrrz, line 16: Would `https://www.youtube.com/watch?list=RDTsYhxMnGYCw` work? To avoid having to pass first video ID.

YoutubeExplode/Playlists/PlaylistClient.cs
```diff
@@ -42,9 +43,10 @@ public class PlaylistClient
         var author = channelId is not null && channelTitle is not null
             ? new Author(channelId, channelTitle)
             : null;
+        //Can't get description from mix playlists
+        var description = playlistExtractor.TryGetPlaylistDescription();
```
Tyrrrz, line 47: Defaulting to empty string is fine here, we were already doing it before in similar cases

YoutubeExplode/Playlists/PlaylistClient.cs
```diff
@@ -67,24 +69,29 @@ public class PlaylistClient
...
     }
 
     /// <summary>
     /// Enumerates batches of videos included in the specified playlist.
     /// </summary>
     public async IAsyncEnumerable<Batch<PlaylistVideo>> GetVideoBatchesAsync(
         PlaylistId playlistId,
-        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+        [EnumeratorCancellation] CancellationToken cancellationToken = default       
+        )
```
Tyrrrz, line 83:
```suggestion
        [EnumeratorCancellation] CancellationToken cancellationToken = default)
```

YoutubeExplode/Playlists/PlaylistClient.cs
```diff
@@ -143,12 +150,20 @@ public class PlaylistClient
                 );
 
                 videos.Add(video);
+
+                
             }
 
+            lastVideoId = playlistExtractor.TryGetLastVideoId();
+            index = playlistExtractor.TryGetLastIndex() ?? 0;
+            visitorData  = visitorData ?? playlistExtractor.TryGetVisitorData();
```
Tyrrrz, line 159: What does `visitorData` actually look like?

YoutubeExplode/Playlists/PlaylistController.cs
```diff
@@ -47,16 +47,65 @@ public PlaylistController(HttpClient http)
...
+        if (!playlist.IsPlaylistDetailsAvailable())
+        {
+            //try getting info from the next endpoint if this one can't get the playlist
+            playlist = await GetPlaylistVideosAsync(playlistId,cancellationToken);
+        }
+
+        return playlist;
+    }
+
+    public async ValueTask<PlaylistExtractor> GetPlaylistVideosAsync(
```
Tyrrrz, line 59: Can we use this for all playlists, including for getting metadata?

Tyrrrz (commented):
Thank you, this is valuable! Left some comments.

YoutubeExplode/Playlists/Playlist.cs
xBaank: It does, but the first video shows as unavailable on the current device. So i think is better to pass the first video ID in the url in this case.

YoutubeExplode/Playlists/PlaylistClient.cs
xBaank: ```"visitorData": "CgtOM25XYWtXNmJBTSis3ISPBg%3D%3D"```
Seems to be like a token, we need to use it  along with the index and lastVideoId if we want to get the next videos on the playlists.

YoutubeExplode/Playlists/PlaylistController.cs
xBaank: I thought we could, but the problem I faced was that it didn't return the playlist description.

YoutubeExplode/Bridge/PlaylistExtractor.cs
xBaank: playlistShareUrl always has this format ```https://www.youtube.com/watch?v=ID&list=ID```, this method does only work with next endpoint responses and next endpoint is used as a fallback of browse enpoint when trying to fetch metadata for mixes playlists.

YoutubeExplode/Playlists/PlaylistClient.cs
xBaank: And visitorData only comes in the first response,  if we didn't use it responses would have all the time the same videos even if we changed the index and lastVideoId (at least for mix playlists)).

YoutubeExplode/Bridge/PlaylistExtractor.cs
Tyrrrz: Do we actually need to extract this considering we already know the first video ID? We can just create the URL ourselves and not rely on `platlistShareUrl` being in the payload.

YoutubeExplode/Bridge/PlaylistExtractor.cs
```diff
@@ -106,60 +141,70 @@ internal partial class PlaylistExtractor
...
+        GetVideos()?
+            .LastOrDefault()?
+            .TryGetVideoId()
+    );
+
+    public int? TryGetLastIndex() => Memo.Cache(this, () =>
+        TryGetPlaylistProperty()?
+            .GetPropertyOrNull("contents")?
+            .EnumerateArrayOrNull()?
+            .Select(j => j.GetPropertyOrNull("playlistPanelVideoRenderer"))
```
Tyrrrz, line 185: This could probably be replaced with `GetVideos` too?

YoutubeExplode/Bridge/PlaylistExtractor.cs
```diff
@@ -106,60 +141,70 @@ internal partial class PlaylistExtractor
...
+            .Select(j => j.GetPropertyOrNull("nextContinuationData"))
             .WhereNotNull()
-            .FirstOrNull()?
-            .GetPropertyOrNull("continuationEndpoint")?
-            .GetPropertyOrNull("continuationCommand")?
-            .GetPropertyOrNull("token")?
+            .FirstOrDefault()
+            .GetPropertyOrNull("continuation")?
+            .GetStringOrNull()
+    );
```
Tyrrrz, line 202: Are we actually passing this anywhere now?

YoutubeExplode/Playlists/Playlist.cs
```diff
@@ -37,13 +39,15 @@ public class Playlist : IPlaylist
         string title,
         Author? author,
         string description,
-        IReadOnlyList<Thumbnail> thumbnails)
+        IReadOnlyList<Thumbnail> thumbnails,
+        string? url = null)
```
Tyrrrz, line 43: We'd have to pass `url` in a separate overload (with and without `url`) due to backwards compatibility

YoutubeExplode/Playlists/PlaylistController.cs
```diff
@@ -47,16 +47,65 @@ public PlaylistController(HttpClient http)
         var raw = await SendHttpRequestAsync(request, cancellationToken);
         var playlist = PlaylistExtractor.Create(raw);
 
-        if (!playlist.IsPlaylistAvailable())
+        if (!playlist.IsPlaylistDetailsAvailable())
+        {
+            //try getting info from the next endpoint if this one can't get the playlist
+            playlist = await GetPlaylistVideosAsync(playlistId,cancellationToken);
+        }
```
Tyrrrz, line 54: Does this ever happen?

xBaank: yes, the playlist id that is used in #555 or this playlists id ```RDMM3Kuw0ZwI0WY``` would get the metadata from next endpoint.

YoutubeExplode/Playlists/Playlist.cs
xBaank: do you mean removing it as an optional parameter and making another constructor overload?

YoutubeExplode/Bridge/PlaylistExtractor.cs
xBaank: Not really, so i could remove this method  as we don't use it anymore

YoutubeExplode/Bridge/PlaylistExtractor.cs
xBaank: Yes, just have to add a method that to ```PlaylistVideoExtractor``` that gets the video index.

YoutubeExplode/Playlists/Playlist.cs
Tyrrrz: > do you mean removing it as an optional parameter and making another constructor overload?

Yeah. That's because other libraries that depend on YoutubeExplode will fail to bind to this constructor since the signature has changed.

YoutubeExplode/Bridge/PlaylistExtractor.cs
Tyrrrz: Yes, please remove any code that is not longer used in this new flow

YoutubeExplode/Bridge/PlaylistExtractor.cs
Tyrrrz: Yeah, let's do that

PR #653, opened 2022-09-08
Use TV embed client to work around age-restricted videos

### LINKED ISSUE #581: Age restricted videos are broken again
### Version

6.0.4 & master

### Details

With the removal of `get_video_info` we currently don't have a way to resolve age-restricted videos anymore.

### Steps to reproduce

```csharp
// Arrange

// Act
var manifest = await youtube.Videos.Streams.GetManifestAsync("SkRSXFQerZs");

// Assert
manifest.Streams.Should().NotBeEmpty();
```

xBaank:
Added TV embed client as a fallback for age restricted videos.

Closes #581


YoutubeExplode/Videos/VideoController.cs
```diff
@@ -42,14 +42,10 @@ public VideoController(HttpClient http)
         );
     }
 
-    public async Task<PlayerResponseExtractor> GetPlayerResponseAsync(
+    public async Task<PlayerResponseExtractor> GetPlayerResponseAndroidClientAsync(
```
Tyrrrz, line 45: Can you use the previous signature, i.e. `GetPlayerResponseAsync(VideoId videoId, bool isEmbedded, CancellationToken cancellationToken = default)` and just get the `TVEMBED` version if `isEmbedded` is `true`? To reduce code churn a bit.

Tyrrrz:
Can you also unskip tests that work with age-restricted videos to make sure everything is in order?

For example, here and in other places (search in files for `Skip`): https://github.com/Tyrrrz/YoutubeExplode/blob/4173ceb7e82dc6d6208f4cbd49456c38b09edc45/YoutubeExplode.Tests/StreamSpecs.cs#L72-L73

YoutubeExplode/Videos/VideoController.cs
xBaank: sure

xBaank:
> Thank you!
> 
> Can you also unskip tests that work with age-restricted videos to make sure everything is in order?
> 
> For example, here and in other places (search in files for `Skip`):
> 
> https://github.com/Tyrrrz/YoutubeExplode/blob/4173ceb7e82dc6d6208f4cbd49456c38b09edc45/YoutubeExplode.Tests/StreamSpecs.cs#L72-L73

https://github.com/yt-dlp/yt-dlp/pull/575#issuecomment-888837000

Would it be ok if I add cases for each tier described here?, I think there are some cases that doesn't work with Tv embed client such as music videos.

Tyrrrz:
> Would it be ok if I add cases for each tier described here?, I think there are some cases that doesn't work with Tv embed client such as music videos.

Sure. Just keep them to a minimal set that's relevant to our case.

xBaank:
After some research I realized that TV embed client could behave different for each video. For example if you make a request for `SkRSXFQerZs` , It responds with a signatureCipher, but when sending the request to other age restricted video like `rXMX4YJ7Lks` or `c0i88t0Kacs`, It responds directly with the url. 
- Should the SignatureScrambling code removed in this commit work for those cases where we need to get the s param? https://github.com/Tyrrrz/YoutubeExplode/commit/4bdbbf23939db45dd19fba396389527fa3cde857
- Do players have fixed url or is there some way of getting it's url like finding it on the video watch page response or an api enpoint?

Tyrrrz:
> * Should the SignatureScrambling code removed in this commit work for those cases where we need to get the s param? [4bdbbf2](https://github.com/Tyrrrz/YoutubeExplode/commit/4bdbbf23939db45dd19fba396389527fa3cde857)

Theoretically yes, but it's possible it was changed since then.

> * Do players have fixed url or is there some way of getting it's url like finding it on the video watch page response or an api enpoint?

They are not fixed, you need to scan the page and look for the respective `<script src="...">`, if I remember correctly.

YoutubeExplode/Videos/VideoController.cs
xBaank: When using the TV embed client we need to send the signatureTimestamp from the player in the request, Do I add another parameter for the signatureTimestamp?

YoutubeExplode/Videos/Streams/StreamClient.cs
```diff
@@ -32,11 +32,32 @@ public StreamClient(HttpClient http)
...
+    private string? TryGetSignatureTimestamp(string playerSource)
+    {
+        var match = Regex.Match(playerSource, @"(?:signatureTimestamp|sts)\s*:\s*(?<sts>[0-9]{5})");
+        return !match.Success ? null : match.Groups[1].Value;
+    }
+
     private async ValueTask PopulateStreamInfosAsync(
         ICollection<IStreamInfo> streamInfos,
         IEnumerable<IStreamInfoExtractor> streamInfoExtractors,
+        string? playerSource,
```
Tyrrrz: Should accept `PlayerSourceExtractor` at this point here, instead of raw source

YoutubeExplode/Videos/Streams/StreamClient.cs
```diff
@@ -32,11 +32,32 @@ public StreamClient(HttpClient http)
...
+    {
+        //Use iframe api to get player version
+        var iframeContent = await _http.GetStringAsync("https://www.youtube.com/iframe_api");
+        
+        var version = Regex.Match(iframeContent, @"player\\?/([0-9a-fA-F]{8})\\?/");
+        if (!version.Success) return null;
+
+        //Download player and return it
+        var source = await _http.GetStringAsync($"https://www.youtube.com/s/player/{version.Groups[1]}/player_ias.vflset/en_US/base.js");
+        return source;
```
Tyrrrz: Are you sure you will always get the same player here and on the watch page?

xBaank: this is the way yt-dlp gets the player url

https://github.com/yt-dlp/yt-dlp/blob/c26f9b991a0681fd3ea548d535919cec1fbbd430/yt_dlp/extractor/youtube.py#L2658-L2666

YoutubeExplode/Videos/Streams/StreamClient.cs
xBaank: I changed it to accept `SignatureScrambler ` as we only need the operations to unscramble.

YoutubeExplode/Videos/Streams/StreamClient.cs
```diff
@@ -31,12 +32,29 @@ public StreamClient(HttpClient http)
...
+
 
     private async ValueTask PopulateStreamInfosAsync(
         ICollection<IStreamInfo> streamInfos,
         IEnumerable<IStreamInfoExtractor> streamInfoExtractors,
+        SignatureScrambler signatureScrambler,
         CancellationToken cancellationToken = default)
     {
+
         foreach (var streamInfoExtractor in streamInfoExtractors)
```
Tyrrrz, line 58:
```suggestion
    private async Task<PlayerSourceExtractor?> TryGetPlayerSourceAsync()
    {
        var iframeContent = await _http.GetStringAsync("https://www.youtube.com/iframe_api");

        var version = Regex.Match(iframeContent, @"player\\?/([0-9a-fA-F]{8})\\?/").Groups[1].Value;
        if (string.IsNullOrWhiteSpace(version))
        	return null;

        var source = await _http.GetStringAsync($"https://www.youtube.com/s/player/{version}/player_ias.vflset/en_US/base.js");

        return PlayerSourceExtractor.Create(source);
    }

    private async ValueTask PopulateStreamInfosAsync(
        ICollection<IStreamInfo> streamInfos,
        IEnumerable<IStreamInfoExtractor> streamInfoExtractors,
        SignatureScrambler signatureScrambler,
        CancellationToken cancellationToken = default)
    {
        foreach (var streamInfoExtractor in streamInfoExtractors)
```

YoutubeExplode/Videos/Streams/StreamClient.cs
```diff
@@ -164,17 +187,28 @@ public StreamClient(HttpClient http)
             );
         }
 
+        var signatureScrambler = SignatureScrambler.Null;
+
+        if (!playerResponse.IsVideoPlayable())
+        {
+            var playerSource = await TryGetPlayerSourceExtractor() ?? throw new YoutubeExplodeException("Could not get player");
```
Tyrrrz, line 194:
```suggestion
            var playerSource = await TryGetPlayerSourceAsync() ?? throw new YoutubeExplodeException("Could not get player");
```

YoutubeExplode/Videos/Streams/StreamClient.cs
```diff
@@ -191,13 +225,13 @@ public StreamClient(HttpClient http)
     }
 
     /// <summary>
-    /// Gets the HTTP Live Stream (HLS) manifest URL for the specified video (if it is a livestream).
+    ///     Gets the HTTP Live Stream (HLS) manifest URL for the specified video (if it is a livestream).
```
Tyrrrz, line 228:
```suggestion
    /// Gets the HTTP Live Stream (HLS) manifest URL for the specified video (if it is a livestream).
```

YoutubeExplode/Videos/Streams/StreamClient.cs
```diff
@@ -46,6 +64,16 @@ public StreamClient(HttpClient http)
...
+            
+            var signatureParameterp = streamInfoExtractor.TryGetSignatureParameter();
+            var signature = streamInfoExtractor.TryGetSignature();
+            
+            //If player source is not null and signature is not null
+            if (signatureParameterp is not null && signature is not null)
+            {
+                var spValue = signatureScrambler.Unscramble(signature);
+                url += $"&{signatureParameterp}={spValue}";
+            }
```
Tyrrrz, line 76:
```suggestion
            var signatureParameter = streamInfoExtractor.TryGetSignatureParameter();
            var signature = streamInfoExtractor.TryGetSignature();

            if (!string.IsNullOrWhiteSpace(signatureParameter) && !string.IsNullOrWhiteSpace(signature))
            {
                var spValue = signatureScrambler.Unscramble(signature);
                url += $"&{signatureParameterp}={spValue}";
            }
```

YoutubeExplode/Videos/Streams/StreamClient.cs
```diff
@@ -46,6 +61,15 @@ public StreamClient(HttpClient http)
...
                 throw new YoutubeExplodeException("Could not extract stream URL.");
+            
+            var signatureParameter = streamInfoExtractor.TryGetSignatureParameter();
+            var signature = streamInfoExtractor.TryGetSignature();
+
+            if (!string.IsNullOrWhiteSpace(signatureParameter) && !string.IsNullOrWhiteSpace(signature))
+            {
+                var spValue = signatureScrambler.Unscramble(signature);
+                url += $"&{signatureParameterp}={spValue}";
+            }
```
Tyrrrz, line 72: This is not exactly how it should work, there should be a default parameter in case `signatureParameter` is not set. If I remember correctly, it was `sig` or `s`. Also, Don't update the URL like that, use `Url.SetQueryParameter(...)`.

xBaank: Is guess this what you mean right? 

https://github.com/Tyrrrz/YoutubeExplode/commit/4bdbbf23939db45dd19fba396389527fa3cde857#diff-87f8ee2ac1480a240600a576eaa5a9c96321a809d9d861d5321975c2d44ca75aL37-L51

Tyrrrz: Yes, exactly 🙂

YoutubeExplode/Videos/Streams/StreamClient.cs
```diff
@@ -46,6 +77,10 @@ public StreamClient(HttpClient http)
             var url =
                 streamInfoExtractor.TryGetUrl() ??
                 throw new YoutubeExplodeException("Could not extract stream URL.");
+            
+            var signatureParameter = streamInfoExtractor.TryGetSignatureParameter();
+            var signature = streamInfoExtractor.TryGetSignature();
+            url = UnscrambleStreamUrl(signatureScrambler, url, signature, signatureParameter);
```
Tyrrrz, line 83:
```suggestion
            var url = UnscrambleStreamUrl(
                streamInfoExtractor.TryGetUrl() ??
                throw new YoutubeExplodeException("Could not extract stream URL."),
                streamInfoExtractor.TryGetSignature(),
                streamInfoExtractor.TryGetSignatureParameter()
            );
```

YoutubeExplode/Bridge/PlayerSourceExtractor.cs
```diff
@@ -0,0 +1,101 @@
...
+            .Groups[0]
+            .Value
+            .NullIfWhiteSpace()
+    );
+    
+    public string? TryGetSignatureTimestamp() => Memo.Cache(this, () =>
+    {
+        var match = Regex.Match(_content, @"(?:signatureTimestamp|sts)\s*:\s*(?<sts>[0-9]{5})");
+        return !match.Success ? null : match.Groups[1].Value;
+    });
```
Tyrrrz, line 27:
```suggestion
    public string? TryGetSignatureTimestamp() => Memo.Cache(this, () =>
				Regex.Match(_content, @"(?:signatureTimestamp|sts)\s*:\s*(?<sts>[0-9]{5})")
        		.Groups[1]
            .Value
            .NullIfWhiteSpace()
    );
```

YoutubeExplode/Bridge/PlayerSourceExtractor.cs
```diff
@@ -0,0 +1,102 @@
...
+        if (string.IsNullOrWhiteSpace(objName))
+            return null;
+
+        var escapedObjName = Regex.Escape(objName);
+
+        return Regex.Match(_content, $@"var\s+{escapedObjName}=\{{(\w+:function\(\w+(,\w+)?\)\{{(.*?)\}}),?\}};",
+                RegexOptions.Singleline)
+            .Groups[0]
+            .Value
+            .NullIfWhiteSpace();
```
Tyrrrz, line 49:
```suggestion
        return Regex.Match(_content, $@"var\s+{Regex.Escape(objName)}=\{{(\w+:function\(\w+(,\w+)?\)\{{(.*?)\}}),?\}};",
                RegexOptions.Singleline)
            .Groups[0]
            .Value
            .NullIfWhiteSpace();
```

YoutubeExplode/Videos/Streams/StreamClient.cs
```diff
@@ -31,10 +32,40 @@ public StreamClient(HttpClient http)
...
+        var iframeContent = await _http.GetStringAsync("https://www.youtube.com/iframe_api");
+
+        var version = Regex.Match(iframeContent, @"player\\?/([0-9a-fA-F]{8})\\?/").Groups[1].Value;
+        if (string.IsNullOrWhiteSpace(version))
+        	return null;
+
+        var source = await _http.GetStringAsync($"https://www.youtube.com/s/player/{version}/player_ias.vflset/en_US/base.js");
+
+        return PlayerSourceExtractor.Create(source);
+    }
```
Tyrrrz, line 47: One last thing, can you move this to `StreamsController` please? I think that's a better place for it

Tyrrrz:
Merged, thanks!

PR #661, opened 2022-10-26
Add Heatmap to Video

drasticactions:
This PR adds "Heatmap" to the `video` object. 

スクリーンショット 2022-10-26 15 01 58

YouTube started to offer a "heatmap" or "Most Replayed" view of (some) videos on their site. While not offered on every video, it can be useful for seeing which sections of a video are the most viewed or scrubbed through. My PR:

- Introduces a new InitialDataExtractor, which takes the InitialData object from the YouTube webpage and parses it.
- Introduces a HeatmapExtractor.
- Introduces Heatmap as a new object.

AFAIK, the only way to get access to this data is from this InitialData object, so I did not add it to the IVideo interface, since I think you can only get it when getting the player stats. I also added a new unit test for check for the existence of the item. If the video doesn't have a heatmap, it is empty rather than null.

### DIFF
--- a/YoutubeExplode.Tests/TestData/VideoIds.cs
+++ b/YoutubeExplode.Tests/TestData/VideoIds.cs
@@ -10,6 +10,7 @@ internal static class VideoIds
     public const string LiveStreamRecording = "rsAAeyAr-9Y";
     public const string ContainsHighQualityStreams = "V5Fsj_sCKdg";
     public const string ContainsDashManifest = "AI7ULzgf8RU";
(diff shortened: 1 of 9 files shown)


Tyrrrz:
Hi. Thanks for the PR, but I'm currently not looking to extend the scope of the library in terms of functionality. You should've really created an issue first to save some effort. Sorry.

drasticactions:
@Tyrrrz No worries, I should have asked first. I can always fork it and keep going, lol. Thanks anyway!

Tyrrrz:
🙂

PR #662, opened 2022-10-26
Set StatusCode on HttpRequestException for .NET 5 and greater

h0lg:
...to spare consumers parsing it out of the message.

Helps with debugging inputs where a `404` has more meaning than other `HttpStatusCode`s.


Tyrrrz:
Thanks!

PR #736, opened 2023-09-02
Fix search feature on mobile devices

### LINKED ISSUE #735: Search feature crashes on mobile devices
### Version

v6.3.3

### Platform

Android

### Steps to reproduce

Executing a simple search query now crashes on mobile devices (Android and most likely iOS as well). You can replicate the issue by downloading an android emulator and running the following code:

```c#
using YoutubeExplode;
using YoutubeExplode.Common;

var videos = await youtube.Search.GetVideosAsync("blender tutorials");
```

### Details

It only crashes (returns an Http 400 Bad Request message) in the latest version (v6.3.3) because of the `NonDisposableHttpContent` class in [HttpExtensions](https://github.com/Tyrrrz/YoutubeExplode/blob/master/YoutubeExplode/Utils/Extensions/HttpExtensions.cs). To be more specific, it would be [this line](https://github.com/Tyrrrz/YoutubeExplode/blob/master/YoutubeExplode/Utils/Extensions/HttpExtensions.cs#L20) in the class that was added. I haven't researched why exactly it crashes on Android but apparently copying a string content instead of copying the stream works.

### Checklist

- [X] I have provided a descriptive title for this issue
- [X] I have made sure that that this bug is reproducible on the latest version of the package
- [X] I have provided all the information needed to reproduce this bug as efficiently as possible

jerry08:
Closes #735

### DIFF
--- a/YoutubeExplode/Utils/Extensions/HttpExtensions.cs
+++ b/YoutubeExplode/Utils/Extensions/HttpExtensions.cs
@@ -1,6 +1,4 @@
-﻿using System.IO;
-using System.Net;
-using System.Net.Http;
+﻿using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -8,32 +6,14 @@ namespace YoutubeExplode.Utils.Extensions;
 
 internal static class HttpExtensions
 {
-    private class NonDisposableHttpContent : HttpContent
-    {
-        private readonly HttpContent _content;
-
-        public NonDisposableHttpContent(HttpContent content) => _content = content;
-
-        protected override async Task SerializeToStreamAsync(
-            Stream stream,
-            TransportContext? context
-        ) => await _content.CopyToAsync(stream);
-
-        protected override bool TryComputeLength(out long length)
-        {
-            length = default;
-            return false;
-        }
-    }
-
-    public static HttpRequestMessage Clone(this HttpRequestMessage request)
+    public static async ValueTask<HttpRequestMessage> CloneAsync(this HttpRequestMessage request)
     {
         var clonedRequest = new HttpRequestMessage(request.Method, request.RequestUri)
         {
             Version = request.Version,
             // Don't dispose the original request's content
             Content = request.Content is not null
-                ? new NonDisposableHttpContent(request.Content)
+                ? new StringContent(await request.Content.ReadAsStringAsync())
                 : null
         };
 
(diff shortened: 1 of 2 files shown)


YoutubeExplode/Utils/Extensions/HttpExtensions.cs
```diff
@@ -1,39 +1,19 @@
...
-
-    public static HttpRequestMessage Clone(this HttpRequestMessage request)
+    public static async ValueTask<HttpRequestMessage> CloneAsync(this HttpRequestMessage request)
     {
         var clonedRequest = new HttpRequestMessage(request.Method, request.RequestUri)
         {
             Version = request.Version,
             // Don't dispose the original request's content
             Content = request.Content is not null
-                ? new NonDisposableHttpContent(request.Content)
```
Tyrrrz, line 36: Would it make more sense to use `new StreamHttpContent(await request.Content.ReadAsStreamAsync())` as a more general solution?

Tyrrrz:
Thanks for the PR! I left a comment. Also please merge `master` in because to fix the tests failing.

YoutubeExplode/Utils/Extensions/HttpExtensions.cs
jerry08: I just tested with `new StreamContent(await request.Content.ReadAsStreamAsync())` and it doesn't work unfortunately. It would return an `Http 400 Bad Request` again.

jerry08:
All done. Thanks for your quick response.

YoutubeExplode/Utils/Extensions/HttpExtensions.cs
Tyrrrz: What about `new ByteArrayContent(await request.Content.ReadAsBytesAsync())`?

jerry08: I tried it and still returns the same error. I will look into detail why using a `ByteArrayContent` or `StreamContent` doesn't work on Android.

Tyrrrz: Yes, please do. It's really weird that only `StringContent` works. Ideally, we should re-encoding the data from/to string because it may be wasteful.

jerry08: I just did some more testing and noticed that initializing a new `StringContent` has default headers for posting strings, such as `Content-Type` along with `CharSet` headers. Using `new StreamContent(...)` doesn't copy headers from `StringContent` so I updated my pull request to copy the headers from the `HttpContent`. I tested it and it works on mobile devices. Without the headers, it throws an exception but not on Desktop.

Tyrrrz: Interesting! Thanks

Issue #762, opened 2024-01-18
Getting `System.Net.CookieException` on some requests


tjvxh:
### Version

6.3.11

### Platform

.NET Framework 4.8

### Steps to reproduce

Download any video (I used https://www.youtube.com/watch?v=i8CDxQc_9-E) and this bug is pretty likely.

I used YoutubeExplode.Videos.Video video = await youtube.Videos.GetAsync(input.Text); My code worked perfectly as of a couple days ago and suddenly I got bugs preventing download of any video.

### Details

I think a new Youtube security feature broke something. I tried older versions of YoutubeExplode and encountered the same error. When trying to download any video I get Exception thrown: 'System.Net.CookieException' in mscorlib.dll.

More exception details:

```
System.Net.CookieException: An error occurred when parsing the Cookie header for Uri 'https://www.youtube.com/watch?v=i8CDxQc_9-E&bpctr=9999999999&hl=en'. ---> System.Net.CookieException: The 'Domain'='m.youtube.com' part of the cookie is invalid.
   at System.Net.Cookie.VerifySetDefaults(CookieVariant variant, Uri uri, Boolean isLocalDomain, String localDomain, Boolean set_default, Boolean isThrow)
   at System.Net.CookieContainer.CookieCutter(Uri uri, String headerName, String setCookieHeader, Boolean isThrow)
   --- End of inner exception stack trace ---
   at System.Net.CookieContainer.CookieCutter(Uri uri, String headerName, String setCookieHeader, Boolean isThrow)
   at System.Net.CookieContainer.SetCookies(Uri uri, String cookieHeader)
   at YoutubeExplode.YoutubeHttpHandler.HandleResponse(HttpResponseMessage response) in /_/YoutubeExplode/YoutubeHttpHandler.cs:line 155
   at YoutubeExplode.YoutubeHttpHandler.<SendAsync>d__5.MoveNext() in /_/YoutubeExplode/YoutubeHttpHandler.cs:line 169
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at PolyfillExtensions.<GetStringAsync>d__4.MoveNext() in /_1/polyshim/1.9.0/contentFiles/cs/any/PolyShim/Net50/HttpClient.cs:line 95
[...]

PR #763, opened 2024-01-18
Fix cookies parse

tqk2811:
Fix https://github.com/Tyrrrz/YoutubeExplode/issues/762


Tyrrrz:
Thanks for the PR, but I ended up going for a simpler fix in #765.

PR #775, opened 2024-02-19
Handle paid videos in search

pvogt:
Added code to ChannelId string to Json parse the Channeld ("browseId") string of a pay-to-watch video such as: https://www.youtube.com/show/SC5q_AUmH7GP_hCdqCc1QQXw
when these type of videos come up in a search query.

Closes #774

PR #776, opened 2024-02-20
Add test case for a playlist that contains paid videos

pvogt:
Related to #775

### DIFF
--- a/YoutubeExplode.Tests/PlaylistSpecs.cs
+++ b/YoutubeExplode.Tests/PlaylistSpecs.cs
@@ -163,6 +163,7 @@ public class PlaylistSpecs(ITestOutputHelper testOutput)
     [InlineData(PlaylistIds.UserUploads)]
     [InlineData(PlaylistIds.ContainsLongVideos)]
     [InlineData(PlaylistIds.Weird)]
+    [InlineData(PlaylistIds.PaidVideo)]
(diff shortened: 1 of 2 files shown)


Tyrrrz:
@pvogt `H4UDZ3MYBJM` does not appear to be a valid playlist ID. Can you share the link?

Also, it would be better if these changes were part of your original PR at #775

pvogt:
> @pvogt `H4UDZ3MYBJM` does not appear to be a valid playlist ID. Can you share the link?
> 
> Also, it would be better if these changes were part of your original PR at #775

@Tyrrrz
https://www.youtube.com/show/SC5q_AUmH7GP_hCdqCc1QQXw is the direct link to the paid videos. If any of these videos came up in the search query using the function:
```
youtube.Search.GetResultsAsync("search query")
```
then the program would error with `Failed to extract the video channel ID.` message.

I was confused as to why `PlaylistIds.cs` and `PlaylistSpecs.cs` need to be modified to check if a Paid Video can be searched without errors.
I figured https://github.com/Tyrrrz/YoutubeExplode/blob/cb35878ddbed3f01297f3e1047cd1f5800dc903d/YoutubeExplode.Tests/VideoSpecs.cs#L94
instead needed to be modified to add something like: `[InlineData(VideoIds.PaidVideo)]` to obtain the metadata such as the `ChannelId`.

PR #784, opened 2024-04-05
Fix quote escapting

xkcd00:
Fixes issue in YoutubeDownloader where quotes are not escaped in search query.


Tyrrrz:
Hi. Thanks for the PR.

Can you move the encoding part here instead: https://github.com/Tyrrrz/YoutubeExplode/blob/4852ebbf581f4b34f3c47e2bd65f7ec47205b9d9/YoutubeExplode/Search/SearchController.cs#L26

Tyrrrz:
Fixed by #785

PR #788, opened 2024-05-17
Search not working with query string in Hebrew fix

### LINKED ISSUE #787: Search not working with query string in Hebrew
### Version

6.3.14

### Platform

.NET 8 / linux

### Steps to reproduce

var videos = await _youtubeClient.Search.GetVideosAsync("נועה קירל") the bug reproduce on any search method

### Details

where searching in version 6.3.13 gets the videos related to the query, with the search query in the newer version the hebrew will transform to something youtube.
i looked into the code and the latest changes and this has direct relation to "Fix search failing when query contains certain characters #785"
my take on a quick google search is that if the search string doesn't include any special char's we might need to send it as and if it does include special char's we need to UrlEncode it.
i can open a pr for a suggested fix please let me know if it ok?

### Checklist

- [X] I have looked through existing issues to make sure that this bug has not been reported before
- [X] I have provided a descriptive title for this issue
- [X] I have made sure that this bug is reproducible on the latest version of the package
- [X] I have provided all the information needed to reproduce this bug as efficiently as possible

razper:
Closes #787

### DIFF
--- a/YoutubeExplode.Tests/SearchSpecs.cs
+++ b/YoutubeExplode.Tests/SearchSpecs.cs
@@ -1,3 +1,4 @@
+using System.Linq;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Xunit;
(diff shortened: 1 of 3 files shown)


Tyrrrz:
Thanks!

PR #801, opened 2024-07-20
Fetch video count in playlist metadata

### LINKED ISSUE #356: How do you get the number of videos in playlist?
How can I get the total number of videos in a playlist?

client.Playlists.GetVideosAsync (id) doesn't seem to support ToListAsync ()

This is necessary because I want to set a max limit since some lists are huge.
If it's a reasonable number, then I show the list.

Can this be added to the Playlist metadata?
Something like this:

   Playlist pll = await client.Playlists.GetAsync (id);
   if (pll.Count > MAX_PCOUNT) {
     return (ERROR_TOO_MANY);
   }

IcySnex:
Closes #356
// Provided solution wasn't actually a solution. Just a workaround.

---

#### Added support for "vidoes count" property for playlist fetching:
Knowing the estimate videos count provided by the offical YouTube API without having to reiterate over the entire video collection is needed (especially when there are hundreds of videos in the playlist)
- Added `int VideosCount` to `Playlist`
- Added `int? VideosCount` to `IPlaylistData`, `PlaylistBrowseResponse`, `PlaylistNextResponse`
- Updated `PlaylistSpecs`-Tests

### DIFF
--- a/YoutubeExplode.Tests/PlaylistSpecs.cs
+++ b/YoutubeExplode.Tests/PlaylistSpecs.cs
@@ -31,6 +31,7 @@ public class PlaylistSpecs(ITestOutputHelper testOutput)
         playlist
             .Description.Should()
             .Contain("Digital Analytics Fundamentals course on Analytics Academy");
+        playlist.VideosCount.Should().Be(22);
         playlist.Thumbnails.Should().NotBeEmpty();
     }
 
@@ -82,6 +83,7 @@ public class PlaylistSpecs(ITestOutputHelper testOutput)
         playlist.Url.Should().NotBeNullOrWhiteSpace();
         playlist.Title.Should().NotBeNullOrWhiteSpace();
         playlist.Description.Should().NotBeNull();
+        playlist.VideosCount.Should().NotBe(null);
         playlist.Thumbnails.Should().NotBeEmpty();
     }
 
--- a/YoutubeExplode.Tests/PlaylistSpecs.cs
+++ b/YoutubeExplode.Tests/PlaylistSpecs.cs
@@ -83,7 +83,6 @@ public class PlaylistSpecs(ITestOutputHelper testOutput)
         playlist.Url.Should().NotBeNullOrWhiteSpace();
         playlist.Title.Should().NotBeNullOrWhiteSpace();
         playlist.Description.Should().NotBeNull();
-        playlist.VideosCount.Should().NotBe(0);
         playlist.Thumbnails.Should().NotBeEmpty();
     }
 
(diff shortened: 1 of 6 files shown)


Tyrrrz:
Thank you. I'm not sure I like that some playlists will have `null` count with this approach.

IcySnex:
During my testing I didn't find any playlists where YouTube doesn't return the number of videos, but for safety reasons I made it nullable internally and set it to 0 in the `PlaylistClient` if it is null, similar to the `Description` property

YoutubeExplode.Tests/PlaylistSpecs.cs
```diff
@@ -82,6 +83,7 @@ public async Task I_can_get_the_metadata_of_any_available_playlist(string playli
         playlist.Url.Should().NotBeNullOrWhiteSpace();
         playlist.Title.Should().NotBeNullOrWhiteSpace();
         playlist.Description.Should().NotBeNull();
+        playlist.VideosCount.Should().NotBe(null);
```
Tyrrrz, line 86: `VideosCount` is always not-`null`, so maybe `Should().NotBe(0)`?

Tyrrrz:
It looks like the `PlaylistNext` response parser is hard-codded to return `null`, and we use that primarily for dynamic playlists (such as mixes) and other playlists that we can't fetch normally.

What video count does it return for the non-`PL` playlists from this list? https://github.com/Tyrrrz/YoutubeExplode/blob/9f1bffdf4888b4f1e572d4e65aecee1d46fb0439/YoutubeExplode.Tests/TestData/PlaylistIds.cs#L5-L14

IcySnex:
I have implemented the parsing of the video count for the `PlaylistNextResponse`. I have tested the non-`PL` playlists with this and
most of them work just fine but there are minor issues with `PlaylistIds.VideoMix` & `PlaylistIds.MusicMix`.

I have inspected the YouTube API response and I have noticed that the `ContentRoot` also provides a property called `isInfinite`. If this property is true (which is the case for these Mix-Playlists) then YouTube won't tell the actual video count.

Thus I have added the `IsInfinite`-property to the `PlaylistNextResponse`. But I'm not quite sure what to do with the `VideoCount`-property in this case. Currently It will be set to `int.MaxValue` but I'm not quite sure about this approach:
https://github.com/Tyrrrz/YoutubeExplode/blob/0e6400acb6a5e07b56c9766d2f3311285a13b662/YoutubeExplode/Bridge/PlaylistNextResponse.cs#L37-L58

Tyrrrz:
Let's make the `VideoCount` property nullable and have `null` indicate an infinite playlist (explain so in the `<remarks>` block).

IcySnex:
Alright that sounds good. I have implemented it and updated the tests.

Tyrrrz:
Thank you :)

Issue #794, opened 2024-06-15
New anti-bot blocker on Youtube prevents to get video manifests (`Sign in to confirm you’re not a bot`)


Darki93:
### Version

6.3.16

### Platform

.NET 8 / Debian

### Steps to reproduce

Query and download many videos for a longer time on the same system with static IP address.

### Details

I am executing YoutubeExplode on a Debian server with static IP address.
Since a few days I am encountering the following exception on the GetManifestAsync method:

> YoutubeExplode.Exceptions.VideoUnplayableException: Video '<ID>' is unplayable. Reason: 'Sign in to confirm you’re not a bot'.
>    at YoutubeExplode.Videos.Streams.StreamClient.GetStreamInfosAsync(VideoId videoId, PlayerResponse playerResponse, CancellationToken cancellationToken) in /_/YoutubeExplode/Videos/Streams/StreamClient.cs:line 212
>    at YoutubeExplode.Videos.Streams.StreamClient.GetStreamInfosAsync(VideoId videoId, CancellationToken cancellationToken) in /_/YoutubeExplode/Videos/Streams/StreamClient.cs:line 276
>    at YoutubeExplode.Videos.Streams.StreamClient.GetManifestAsync(VideoId videoId, CancellationToken cancellationToken) in /_/YoutubeExplode/Videos/Streams/StreamClient.cs:line 292

Seems like Youtube added a new mechanism to stop multiple videos to be queried/downloaded from the same machine/IP without logging in.

### Checklist

- [X] I have looked through existing issues to make sure that this bug has not been reported before
- [X] I have provided a descriptive title for this issue
- [X] I have made sure that this bug is reproducible on the latest version of the package
- [X] I have provided all the information needed to reproduce this bug as efficiently as possible
- [ ] I have sponsored this project
- [ ] I have not read any of the above and just checked all the boxes to submit the issue

Tyrrrz:
> Seems like Youtube added a new mechanism to stop multiple videos to be queried/downloaded from the same machine/IP without logging in.

I don't think there's much that you can do if that's the case. Would have to sign in.

ReenigneArcher:
This was kind of random before... it's extremely repeatable now.

rpm61:
I am also getting "Video 'AfWnMPDM0G4' is not available." from our Azure servers, noit getting it locally. Is it possible to use proxies with Explode?

Darki93:
[...]

PR #813, opened 2024-09-18
Try adding delay between requests to avoid bot detection

Tyrrrz:
Related to #794


Tyrrrz:
Closing. Adding delay & locking helped with local testing, but not with tests running on public GitHub Actions runners.

PR #843, opened 2024-11-25
Fix playlist search response extraction

h0lg:
because test is red and search is broken on the current master

### DIFF
--- a/YoutubeExplode/Bridge/SearchResponse.cs
+++ b/YoutubeExplode/Bridge/SearchResponse.cs
@@ -30,7 +30,13 @@ internal partial class SearchResponse(JsonElement content)
         ContentRoot
             ?.EnumerateDescendantProperties("lockupViewModel")
             .Select(j => new PlaylistData(j))
-            .ToArray() ?? [];
+            .ToArray()
+        // fall back to older working paths
+        ?? ContentRoot
+            ?.EnumerateDescendantProperties("playlistRenderer")
+            .Select(j => new PlaylistData(j))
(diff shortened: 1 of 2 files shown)


YoutubeExplode/Bridge/SearchResponse.cs
```diff
@@ -129,44 +129,48 @@ internal partial class SearchResponse
...
             AuthorDetails
-                ?.GetPropertyOrNull("navigationEndpoint")
+                ?.GetPropertyOrNull("commandRuns")
+                ?.EnumerateArrayOrNull()
+                ?.ElementAtOrNull(0)
+                ?.GetPropertyOrNull("onTap")
+                ?.GetPropertyOrNull("innertubeCommand")
                 ?.GetPropertyOrNull("browseEndpoint")
                 ?.GetPropertyOrNull("browseId")
                 ?.GetStringOrNull();
```
Tyrrrz, line 162: Please keep the original codepaths but add your new alternatives as a fallback, via `??`. Most YouTube changes are rolled out in waves so both codepaths are likely to be in use at the same time.

h0lg: @Tyrrrz Can do - but wouldn't it make more sense (performance-wise) to **pre**pend the new path alternatives instead of appending them? That way you don't check all the (probably broken) old paths before hitting a working one.

Tyrrrz: Yes, it would make more sense that way, thank you

Issue #853, opened 2024-12-22
 403 (Forbidden) on any video


worbarry:
### Version

6.5.0

### Platform

.NET 8

### Steps to reproduce

static class Program
    {
        static async Task Main(string[] args)
        {
            Console.WriteLine("Youtube VideoDownloader");
            Console.WriteLine("");
            Console.Write("Enter URL: ");
            var videoUrl = Console.ReadLine();

            if (videoUrl != null)
            {
                await Download(videoUrl);
            }
            
        }

        private static async Task Download(string videoUrl)
        {
            var youtube = new YoutubeClient();
            
            var video = await youtube.Videos.GetAsync(videoUrl);
            var streamManifest = await youtube.Videos.Streams.GetManifestAsync(videoUrl);

            var streamInfo = streamManifest
                .GetVideoOnlyStreams()
                .Where(s => s.Container == Container.Mp4)
                .GetWithHighestVideoQuality();
            IProgress<double> progress = new Progress<double>(p => Console.Write($"Progress updated: {p}"));
            await youtube.Videos.Streams.DownloadAsync(streamInfo, $"video.{streamInfo.Container}", progress);
            
        }
    }

### Details

Download video file

result
Unhandled exception. System.Net.Http.HttpRequestException: Response status code does not indicate success: 403 (Forbidden).

### Checklist

- [X] I have looked through existing issues to make sure that this bug has not been reported before
- [X] I have provided a descriptive title for this issue
- [X] I have made sure that this bug is reproducible on the latest version of the package
- [X] I have provided all the information needed to reproduce this bug as efficiently as possible
- [ ] I have sponsored this project
- [ ] I have not read any of the above and just checked all the boxes to submit the issue

Tyrrrz:
What video URL were you using?

worbarry:
I was trying any random video

dsimonubb:
I get now also the 403 Firbidden error when calling 
var streamManifest = await youtube.Videos.Streams.GetManifestAsync(video.Id); on all the videos I try.
I use also version 6.5.0.
A few days ago it still worked.

udi12:
hi there 
been trying this now, and getting "forbidden" on any video...

mohanadshaban:
Same here...

udi12:
Is this happening for every1 or is just my IP blocked?

udi12:
[...]

PR #901, opened 2025-08-07
Ignore 403 Forbidden streams when getting manifest

ZmorzynskiK:
Related to problems in #853

Added ignoring for 403 Forbidden streams when getting manifests.

### DIFF
--- a/YoutubeExplode/Videos/Streams/StreamClient.cs
+++ b/YoutubeExplode/Videos/Streams/StreamClient.cs
@@ -58,7 +58,10 @@ public class StreamClient(HttpClient http)
             contentLength = response.Content.Headers.ContentLength;
 
             // 404 error indicates that the stream is not available
-            if (response.StatusCode == HttpStatusCode.NotFound)
+            if (
+                response.StatusCode == HttpStatusCode.NotFound
+                || response.StatusCode == HttpStatusCode.Forbidden
+            )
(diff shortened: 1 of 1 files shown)


Tyrrrz:
Hi @ZmorzynskiK.
Did you get a chance to test if it significantly affects how many streams are returned?

Tyrrrz:
Unfortunately, this method ends up skipping all streams in the manifest

ZmorzynskiK:
I didn't notice that it skips all the streams. It might skip some streams, but some streams are downloaded (at least we get something, instead of getting nothing, right? :) ).

I sometimes get the `YoutubeExplode.Exceptions.YoutubeExplodeException: Failed to extract the cipher manifest.`  exception, but I guess it's related to some YouTube's algorithm to "fight" the libraries.
I suspect that after ignoring 403 status (instead of finishing early and completely stopping) and moving further with other streams, YT sometimes detects it and hence the exception. But running the same URL a couple of hours later - the error is gone.

Tyrrrz:
It does skip all streams since all of them report 403.

What ends up happening is that it triggers the fallback resolving method that relies on deciphering. Unfortunately, that had its own issues even prior to now, but lately it fails with `Failed to extract the cipher manifest` as you experienced.

In any case, I don't think skipping is the right approach here.

PR #919, opened 2025-10-23
Change the client to avoid the 403 error

pictos:
While look into www I found that issue is related to google now require a POToken, or something like that, and while I didn't find a way to generate/grab/use this token I found a PR that provides a workaround for that issue. Looks like there's one client that doesn't need that token (yet).

- https://github.com/yt-dlp/yt-dlp/pull/14693/files#diff-443dcfbd960da6a3264f7d52869636689f41f36daa876ade16a050c86c542ccdR223

The downside, that I found in my tests, is that the number of results for Muxed (video + audio) are reduced, but you know is better than nothing.

Related to #853

### DIFF
--- a/YoutubeExplode/Videos/VideoController.cs
+++ b/YoutubeExplode/Videos/VideoController.cs
@@ -172,10 +172,8 @@ internal class VideoController(HttpClient http)
               "videoId": {{Json.Serialize(videoId)}},
               "context": {
                 "client": {
-                  "clientName": "ANDROID",
-                  "clientVersion": "20.10.38",
-                  "osName": "Android",
-                  "osVersion": "11",
+                  "clientName": "TVHTML5_SIMPLY_EMBEDDED_PLAYER",
+                  "clientVersion": "2.0",
                   "visitorData": {{Json.Serialize(visitorData)}},
                   "hl": "en",
                   "gl": "US",


Tyrrrz:
Hi, thanks for the PR.

> The downside, that I found in my tests, is that the number of results for Muxed (video + audio) are reduced, but you know is better than nothing.

Weren't we already not getting any muxed streams anyway, for a while?

YoutubeExplode/Videos/VideoController.cs
```diff
@@ -174,8 +172,10 @@ public async ValueTask<PlayerResponse> GetPlayerResponseAsync(
               "videoId": {{Json.Serialize(videoId)}},
               "context": {
                 "client": {
-                  "clientName": "TVHTML5_SIMPLY_EMBEDDED_PLAYER",
-                  "clientVersion": "2.0",
+                  "clientName": "ANDROID",
+                  "clientVersion": "20.10.38",
+                  "osName": "Android",
+                  "osVersion": "11",
```
Tyrrrz: This needs to remain the same because it needs the `TVHTML5_SIMPLY_EMBEDDED_PLAYER` to work around age-restricted videos

pictos: done

Tyrrrz:
Thanks a lot @pictos!

YumYumFisch:
Awesome! Thank you very much for fixing the 403 Forbidden Error 🥳

PR #922, opened 2025-10-26
Get the batch of working playlist videos and stop on the unavailable batch

### LINKED ISSUE #921: Can't fetch a public playlist
### Version

v6.5.6

### Platform

.NET 9.0 / Windows 11

### Steps to reproduce

```csharp
var pl = await youtube.Playlists.GetAsync("youtube.com/playlist?list=PLz-YZsjoSww4AZGEeuJ2BwJ984_TCI0P1");
```

Image

### Details

Expected: playlist metadata

Actual: playlist unavailable

### Checklist

- [x] I have sponsored this project

HossamSaberX:
Partially fixes #921 by handling unavailable playlist batches gracefully.

**Before:** Fetching crashed with `PlaylistUnavailableException` on any unavailable batch.

**After:** Catches the exception to stop enumeration without crashing. 

Note: In practice, unavailable batches omit all videos, so this prevents crashes but may still stop early.
We may think of further improvements (like jumping over batches) can extend coverage but also may miss smth before if the jumping to indices is too large.

### DIFF
--- a/YoutubeExplode/Playlists/PlaylistClient.cs
+++ b/YoutubeExplode/Playlists/PlaylistClient.cs
@@ -4,6 +4,7 @@ using System.Net.Http;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
+using YoutubeExplode.Bridge;
 using YoutubeExplode.Common;
 using YoutubeExplode.Exceptions;
 using YoutubeExplode.Videos;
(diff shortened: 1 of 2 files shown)


YoutubeExplode/Playlists/PlaylistClient.cs
```diff
@@ -83,13 +84,22 @@ public async IAsyncEnumerable<Batch<PlaylistVideo>> GetVideoBatchesAsync(
...
+                    lastVideoId,
+                    lastVideoIndex,
+                    visitorData,
+                    cancellationToken
+                );
+            }
+            catch (PlaylistUnavailableException)
+            {
+                // Stop enumeration if the playlist becomes unavailable
+                yield break;
```
Tyrrrz: Shouldn't we try the next batch still, until it's empty?

PR #927, opened 2025-11-11
Improve MP3 conversion quality

### LINKED ISSUE #926: MP3 quality degraded
### Version

6.5.6

### Platform

.NET 9.0 / Windows 10

### Steps to reproduce

Currently when using MP3 containers, the downloaded file is a constant bitrate of, often, 128 or 160kbps. This was implemented as a fix to #666.

### Details

Detailed explanation in [this comment](https://github.com/Tyrrrz/YoutubeDownloader/issues/23#issuecomment-3507483169)

While the fix for #666 really improves the quality of the output MP3 file, the result is a  CBR file, assuming the source codec's bitrate relates to the MP3 (libmp3codec)'s one. I believe this is, likely most of the times, not true.

To achieve better quality MP3 files, `ffmpeg` could use variable bitrate (VBR) at "best quality", which allows it to set the bitrate to whatever is needed to minimize the loss during the conversion.

The fix itself would be super simple: simply change the passed argument in [@6.5.6:YoutubeExplode.Converter/Converter.cs:103-104](https://github.com/Tyrrrz/YoutubeExplode/blob/6.5.6/YoutubeExplode.Converter/Converter.cs#L103-L104) from `-b:a:<stream> <inferred_bitrate>` to `-q:a:<stream> 0`, which means "best quality variable bitrate".

These flags are documented in [VBR Encoding - Encode/MP3 - FFmpeg](https://trac.ffmpeg.org/wiki/Encode/MP3#VBREncoding).

This way, even if a low quality audio stream is input, `ffmpeg` will do whatever it takes to closely match it to the output MP3's variable bitrate stream. Effectively, lower quality/bitrate input means lower bitrate output, even with "maximum quality".

I couldn't reproduce the original issue discussed in the related SuperUser page, so it really seems worth a shot.

With the fix above I am instantly getting MP3 output in higher bitrate. Some tests I did:

- Outside of YoutubeDownloader / YoutubeExploder, converted 320kbps and 128 kbps mp3s (after normalization) using `-q:a 0` to different results, lower average bitrate to lower bitrate source MP3s.
- Inside YoutubeDownloader, using youtube's opus-251 (details in the linked comment above), bitrates settle in considerably higher values with the new flag, but not necessarily always 260kbps+, depending on the source video it sits around 192kbps or less. It really depends on the actual complexity of the audio stream. (complexity: not quality, necessarily!)

By what I've seen, libmp3codec handles very well the encoding, but there may be reasons people want CBR; I doubt for the today's standards many people would worry about ABR ([average bitrate](https://en.wikipedia.org/wiki/Average_bitrate)).

### Checklist

avengerx:
This closes #926 by using `-q:a:<stream> 0` instead of `-b:a:<stream> <input_stream_bitrate>`

Notice a "drawback" of this approach is, output mp3 files will have considerably higher size. For the "not-audiophile" people and large libraries, this might mean a considerable impact.

The expected average increase is of 2/3rs (66%), but it depends both on the input quality and input codec. Opus in particular should always mean bigger files as it usually packs better quality in smaller bitrates in comparison to lame/mp3.

Ideally YoutubeExplode should accept quality control on the MP3 output, but that's really falling into a "new feature" category and is probably not the focus anymore. Having the best quality possible leaves room for manually reducing the quality after the download if size matters over quality.

p.s.: it was not my decision to join the lines, Visual Studio done it on its own will. Shall we bend to IA's will this time? :)

### DIFF
--- a/YoutubeExplode.Converter/Converter.cs
+++ b/YoutubeExplode.Converter/Converter.cs
@@ -91,7 +91,9 @@ internal partial class Converter(VideoClient videoClient, FFmpeg ffmpeg, Convers
 
         // MP3: explicitly specify the bitrate for audio streams, otherwise their metadata
         // might contain invalid total duration.
-        // https://superuser.com/a/893044
+// MP3: explicitly specify the quality for audio streams, otherwise their metadata
+// might contain invalid total duration.
+// https://superuser.com/a/893044
         if (container == Container.Mp3)
         {
             var lastAudioStreamIndex = 0;


Duc558:
file Hdo56B1YxT4.opus
Size 5.60 Mb
Bitrate; 143,17Kbps
Frequency: 48000 Hz

ffmpeg -y -threads 3 -i "Hdo56B1YxT4.opus" -vn -ar 48000 **-b:a** 128k "Hdo56B1YxT4.mp3"
result:
Size: 5.36 Mb
Bitrate: 128 Kbps
Frequency 48000 Hz

ffmpeg -y -threads 3 -i "Hdo56B1YxT4.opus" -vn -ar 48000 **-q:a** 128k "Hdo56B1YxT4.mp3"
result:
Size: 2.75 Mb
Bitrate: 65 Kbps
Frequency 48000 Hz

Of which, **-b:a** is closer to opus than **-q:a** for the same processing. Of course, with **-b:a** the size is larger but the quality is better.

avengerx:
`-q:a 128k` probably is interpreted as -q:a 9, which is least quality (fastest). You don't specify a bitrate to `-q`, it is based in variable bitrate and is related to "quality", accepting values 0 thru 9, being 0 best quality and 9 worst.
Ref: https://trac.ffmpeg.org/wiki/Encode/MP3#VBREncoding

Tyrrrz:
Thanks 🙏🏻

Tyrrrz:
@avengerx I thought about it after merging, but do you think we could just do `-q:a 0` once, instead of enumerating all streams with `-q:a:[i] 0`? If yes, I'll make the necessary changes.

avengerx:
> @avengerx I thought about it after merging, but do you think we could just do `-q:a 0` once, instead of enumerating all streams with `-q:a:[i] 0`? If yes, I'll make the necessary changes.

I wouldn't mind either way. Perhaps doing it in one shot would be better readability-wise, but I'd say it's more a matter of taste than anything. I can't tell a strong disadvantage either way. Well aside of your spent energy hitting the keyboard keys to make the new commit and pull request. :)
**Edit:** oops, I get what you mean. I really didn't dig down to see if the second parameter to `-q:a:stream` actually works. It made sense that it was used to select a stream from the video, but in my tests I just used "music videos" which aren't multilanguage or anything. It would need more insight about the input and output to tell if that works. And more problems will show up: are there various audio streams (multi-language vid) does YoutubeExplode output one mp3 per audio stream? If it's a run of ffmpeg per audio stream, there may be little benefit from specifying the stream as it is now. If ffmpeg does the coherent output by itself, then maybe, yes...

Well, in short, I don't know! :P It might just work and save that loop!

PR #939, opened 2026-02-25
Handle content length being null in StreamClient

mskovacic:
Allow downloading videos of unknown content length. That is necessary in scenarios when we the videos are being streamed.

Closes #ISSUE_NUMBER


Tyrrrz:
Have you tested it? This still wouldn't be enough to be able to download live videos.

mskovacic:
> Have you tested it? This still wouldn't be enough to be able to download live videos.

I don't know specifically about live videos, but it solved my problem with this  https://www.youtube.com/watch?v=BqvQwOPE3n4

PR #952, opened 2026-05-21
Advanced search filters. Custom language & location params

speige:


YoutubeExplode/Search/SearchClient.cs
```diff
@@ -237,30 +241,27 @@ public async IAsyncEnumerable<Batch<ISearchResult>> GetResultBatchesAsync(
...
-    ) => GetResultBatchesAsync(searchQuery, SearchFilter.None, cancellationToken);
-
     /// <summary>
     /// Enumerates search results returned by the specified query.
     /// </summary>
     public IAsyncEnumerable<ISearchResult> GetResultsAsync(
         string searchQuery,
+        SearchFilter searchFilter = default,
+        string hl = "en",
+        string gl = "US",
```
Tyrrrz, line 251: Does localization affect only how the results are presented, or does it filter by language?

If it filters too, then it should probably be included within `SearchFilter`.

If it's just for presetation, then it would make sense to configure it on the `*Client` object, since it can probably be relevant to all requests.

YoutubeExplode/Search/SearchFilter.cs
```diff
@@ -1,27 +1,279 @@
...
+using System.Collections.Generic;
+
 namespace YoutubeExplode.Search;
 
 /// <summary>
 /// Filter applied to a YouTube search query.
 /// </summary>
-public enum SearchFilter
+public readonly record struct SearchFilter(
+    SearchFilterType? Type = null,
```
Tyrrrz, line 10: `SearchFilterType? Type` -> `SearchFilterKind? Kind`, so as to not confuse with the CLR's concept of "type"

YoutubeExplode/Search/SearchClient.cs
speige: It affects the search results, which is useful if you actually want a spanish video, so maybe SearchFilter would be better.

YoutubeExplode/Search/SearchFilter.cs
speige: Good point. I just threw this together quickly with AI to solve my need, it's working, but I didn't review the code. Feel free to refactor to your preferred style if you have time, or reject/delete the PR if it's too messy.

Tyrrrz:
Closing as stale/unattended

Issue #964, opened 2026-08-07
Closed caption tracks are not retrieved


jickler-o:
### Version

6.6.1

### Environment

Windows 11

### Steps to reproduce

1. Try to get track manifest for either video(which have captions)
    - "https://www.youtube.com/watch?v=k68rAwize_Y"
    - "https://www.youtube.com/watch?v=3LQCDt1dZWA"
2. No closed caption tracks are retrieved.

### Details

These two videos have caption tracks, but the SDK is unable to retrieve one for either video.

### Checklist

- [x] I have looked through existing issues to make sure that this bug has not been reported before
- [x] I have provided a descriptive title for this issue
- [x] I have made sure that this bug is reproducible on the latest version of the package
- [x] I have provided all the information needed to reproduce this bug as efficiently as possible
- [ ] I have sponsored this project

mysteryx93:
This is also fixed by my PR

h0lg:
+1 I see the same issue, no matter the video.

Tyrrrz:
Does this happen when you're authenticated or unauthenticated?

Issue #962, opened 2026-08-03
Not work with https://www.youtube.com/watch?v=5SRLlJ35r3E


khmmarket:
### Version

6.6 Latest @github-actions github-actions released this Apr 22  6.6  79aee02 

### Environment

.NET 10

### Steps to reproduce

Not work with https://www.youtube.com/watch?v=5SRLlJ35r3E

### Details

Not work with https://www.youtube.com/watch?v=5SRLlJ35r3E

### Checklist

- [x] I have looked through existing issues to make sure that this bug has not been reported before
- [x] I have provided a descriptive title for this issue
- [x] I have made sure that this bug is reproducible on the latest version of the package
- [x] I have provided all the information needed to reproduce this bug as efficiently as possible
- [ ] I have sponsored this project

mysteryx93:
Ya it crashes with "Video is not available". Having the same issue with
https://www.youtube.com/watch?v=u9Dg-g7t2l4

mysteryx93:
Issue fixed in this PR
https://github.com/Tyrrrz/YoutubeExplode/pull/965

PR #965, opened 2026-08-19
Fix download errors by using mostly VISIONOS agent

### LINKED ISSUE #902: Failed to extract cipher manifest
### Version

6.5.4

### Platform

.NET 8 / Windows 11

### Steps to reproduce

```
var client = new YoutubeClient();
VideoId id = new("https://www.youtube.com/watch?v=UFdO3wodIJk");
try { await client.Videos.GetAsync(id); }
catch (Exception ex) { Console.WriteLine(ex); } // VideoUnavailableException
try { var cc = await client.Videos.ClosedCaptions.GetManifestAsync(id); }
catch (Exception ex) { Console.WriteLine(ex); } // VideoUnavailableException
try { var streams = await client.Videos.Streams.GetManifestAsync(id); }
catch (Exception ex) { Console.WriteLine(ex); } // YouTubeExplodeException "Failed to extract the cipher manifest."
```

### Details

Thanks for building and maintaining this great library!

While the library works great for most videos I try it with, this particular one fails reliably for me even though on the front-end I don't see anything special about it.

### Checklist

mysteryx93:
Closes #902 #962 #964

The downloader has been broken for some time, and that's the kind of issues that are extremely time-consuming to diagnose.

I have put Grok Build to the task, so this is AI-generated, and I have manually reviewed every line modified. All tests are now passing. 2 were still failing: Live Stream and Upscaler, because of outdated videos. The code is fine. I updated the video IDs. All tests now pass. So far Grok passed every test I gave it; it just keeps going until the tests pass.

### Summary of changes

The failure was not the video. YouTube’s current ANDROID_VR 1.60.19 client (what YoutubeExplode still used) now returns LOGIN_REQUIRED / “Sign in to confirm you’re not a bot” for public videos, including u9Dg-g7t2l4. That makes the main cipher-less path fail, then the #902 cipher fallback also fails because player JS is no longer parseable.

yt-dlp already moved off that client. Its JS-less default is now VISIONOS, which still returns progressive URLs without a PO token.

What changed

1. Primary client: VISIONOS 1.02 — same client yt-dlp uses without Deno. No signature deciphering, and stream URLs actually work (HEAD 200).
2. Fallback: ANDROID 21.26.364 — for “made for kids” videos, which VISIONOS still cannot play. ANDROID still returns muxed itag 18.
3. SABR streams are skipped instead of aborting the whole manifest (Failed to extract the stream URL).

Still broken (unchanged YouTube limits)
• Age-restricted videos still need the cipher client. Player JS is still unparseable.


mysteryx93:
To clarify: the agent fallback works for age-restricted.
Kids videos still fail. These are 2 separate cases.

YoutubeExplode.Tests/StreamSpecs.cs
```diff
@@ -251,6 +255,21 @@ public async Task I_can_download_a_specific_stream_of_a_video(string videoId)
...
+        using var youtube = new YoutubeClient();
+
+        // Act
+        var manifest = await youtube.Videos.Streams.GetManifestAsync(VideoIds.AndroidVrBotCheck);
+
+        // Assert
+        manifest.Streams.Should().NotBeEmpty();
+        manifest.GetAudioStreams().Should().NotBeEmpty();
+        manifest.GetVideoStreams().Should().NotBeEmpty();
+    }
```
Tyrrrz: No need to test this separately

YoutubeExplode.Tests/TestData/VideoIds.cs
```diff
@@ -22,5 +22,7 @@ internal static class VideoIds
     public const string WithClosedCaptions = "YltHGKX80Y8";
     public const string WithBrokenClosedCaptions = "1VKIIw05JnE";
     public const string WithMultipleAudioLanguages = "ngqcjXfggHQ";
-    public const string WithUpscaledStreams = "IFACrIx5SZ0";
+    public const string WithUpscaledStreams = "0N1_0SUGlDQ";
+    public const string AndroidVrBotCheck = "u9Dg-g7t2l4";
```
Tyrrrz: Let's remove this

YoutubeExplode.Tests/TestData/VideoIds.cs
```diff
@@ -22,5 +22,7 @@ internal static class VideoIds
...
+    public const string MadeForKids = "nQuzyJ-C1Fc";
```
Tyrrrz: Let's rename this to just `ForKids` and move it below `ContentCheck*` videos

YoutubeExplode/Videos/VideoController.cs
```diff
@@ -85,67 +85,102 @@ await Http.GetStringAsync(
         }
     }
 
-    public async ValueTask<PlayerResponse> GetPlayerResponseAsync(
-        VideoId videoId,
+    private async ValueTask<PlayerResponse> SendPlayerRequestAsync(
```
Tyrrrz: IMO, this should be inlined. The indirection causes more cognitive overhead than duplication.

YoutubeExplode/Videos/VideoController.cs
```diff
@@ -85,67 +85,102 @@ await Http.GetStringAsync(
...
-        request.Content = new StringContent(
+        request.Content = new StringContent(content);
+        request.Headers.Add("User-Agent", userAgent);
+
+        using var response = await Http.SendAsync(request, cancellationToken);
+        response.EnsureSuccessStatusCode();
+        return PlayerResponse.Parse(await response.Content.ReadAsStringAsync(cancellationToken));
+    }
+
+    public async ValueTask<PlayerResponse> GetPlayerResponseAsync(
```
Tyrrrz, line 124: This could probably be refactored into `GetPlayerResponseForVisionOsAsync` and `GetPlayerResponseForAndroidAsync`. `GetPlayerResponseAsync` would then just handle retrying.

Tyrrrz (changes requested):

mysteryx93:
Do you want me to edit the PR or you handle it from here?

Tyrrrz:
@mysteryx93 yes, please fix as much as you can youself, I'll clean up afterward

YoutubeExplode/Videos/VideoController.cs
mysteryx93: What do you mean by "inlined"? I also saw an issue in your code
`using var request = new HttpRequestMessage(`
This is not recommended and can lead to connection pool saturation. Can be fine for a desktop app; but if used on a web server, this will saturate the connection pool. This is perhaps a different refactor out of this scope though.

mysteryx93: Sending a new request is definitely an issue because you issue many connections to YouTube to analyze each stream one by one. The connection needs to be re-negotiated for every request and it gets stored in the connection pool. Replacing that with HttpClient can make a considerable performance and stability difference.

mysteryx93:
I submitted the refactoring changes. It does considerably improve the code.

2 potentially major issues to look at before deploying.

The TVHTML5 code route is not being used at all; I tried all the unit tests and none hit that route. The new pathway was designed to handle the ForKids issue as well, making this one obsolete.
```
// The only client that can handle age-restricted videos without authentication is the
// TVHTML5_SIMPLY_EMBEDDED_PLAYER client.
// This client does require signature deciphering, so we only use it as a fallback.
```

Furthermore, looking at where it is called: `StreamClient.GetStreamInfosAsync`. We now have 2 separate places to handle fallback logic, which is definitely not ideal.

The whole "Retry with deciphering" fallback could perhaps be removed... but perhaps it would serve in the future too, so it's not an easy call.

mysteryx93:
btw I made some comment about HttpRequestMessage; never mind you're using HttpClient, everything is fine there.

Tyrrrz:
> The TVHTML5 code route is not being used at all; I tried all the unit tests and none hit that route. The new pathway was designed to handle the ForKids issue as well, making this one obsolete.
> 
> ```
> // The only client that can handle age-restricted videos without authentication is the
> // TVHTML5_SIMPLY_EMBEDDED_PLAYER client.
> // This client does require signature deciphering, so we only use it as a fallback.
> ```
> 
> Furthermore, looking at where it is called: `StreamClient.GetStreamInfosAsync`. We now have 2 separate places to handle fallback logic, which is definitely not ideal.
> 
> The whole "Retry with deciphering" fallback could perhaps be removed... but perhaps it would serve in the future too, so it's not an easy call.

@mysteryx93 the TVHTML5 route was used for age-restricted videos, usually marked as such because of violence and/or sexual themes. So it's a separate issue.

mysteryx93:
ah and those 2 tests are set as "Skip" so they weren't run. Those tests fail with "Failed to extract the cipher manifest"; which I suppose is why you disabled the tests. But they work in practice?

Still. We have the retry logic split into 2 areas. It's not ideal but not really an issue either. Considering the 2nd retry path with cipher is a considerably more different/complicated route.

Tyrrrz (approved):
Thanks!

mysteryx93:
@Tyrrrz what's your thoughts about using an extra dependency to support cookies and YouTube Premium streams, with the same method yt-dlp is using?

Do you plan to take care of that eventually? Do you want to keep it fully dependency-free?

Or would you like me to generate the bulk with AI then we refactor it proper, copying yt-dlp's proven path? BUT keeping the dependency fully optional.

As of right now, it seems the cookie-authentication path is useless without proper deciphering support, which can only be done with a real JavaScript engine.

